PR #10, opened 2026-05-01
Add claude-ollama kit

alakae:
# Summary

Adds a `claude-ollama` kit that routes all Claude Code API calls to a local [Ollama](https://ollama.com) instance instead of Anthropic's API. Useful for offline development, cost-free experimentation, and testing with custom local models.

# Spec choices worth flagging for review

- Uses `initFiles` to inject the wrapper script inline
- `ANTHROPIC_` variables are set similar to approach used in [Ollama's own Claude launcher](https://github.com/ollama/ollama/blob/8f39fff70bac0bef2370a6af7020efa29a6a7cad/cmd/launch/claude.go); skips the Claude Code sign-in screen.
- `allowedDomains` is narrow (`localhost:11434` only) — no Anthropic API domains are reachable, which is the whole point.

# Test plan

- `sbx kit validate ./claude-ollama/` ✅
- `go test -v -count=1 -timeout 10m ./claude-ollama/...` ✅
- `sbx run --kit ./claude-ollama/ claude` end-to-end with Ollama running locally ✅

# Origin

Adapted from [`alakae/sbx-kits`](https://github.com/alakae/sbx-kits) (MIT), which was itself adapted from [`dvdksn/kits-cookbook`](https://github.com/dvdksn/kits-cookbook) (claude-safe kit) (also MIT).

### DIFF
--- a/claude-ollama/README.md
+++ b/claude-ollama/README.md
@@ -1,0 +1,45 @@
+# claude-ollama
+
+> `ollama launch claude --model gemma4:e4b-it-q4_K_M` — but in `sbx`.
+
+A fork of the built-in `claude` agent that routes all API calls to a local
+**[Ollama](https://ollama.com)** instance instead of Anthropic's API. Useful for
+offline development, cost-free experimentation, or testing with custom local models.
+
+> **Prerequisite:** Ollama must be running on your host machine at its default port
+> (`localhost:11434`) before starting this sandbox.
+
+## Usage
+
+```console
+$ sbx run claude-ollama --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=claude-ollama" ~/my-project
+```
+
+The agent name passed to `sbx run` (`claude-ollama`) matches the `name:` field in
+the kit's `spec.yaml`.
+
+The default model is `gemma4:e4b-it-q4_K_M`. To use a different model, fork this
+kit and change the `CLAUDE_OLLAMA_MODEL` default in `spec.yaml`.
+
+## What changed vs the built-in `claude`
+
+Instead of calling `api.anthropic.com`, a wrapper script replaces the entrypoint:
+
+```diff
+-  entrypoint:
+-    run: [claude, "--dangerously-skip-permissions"]
++  entrypoint:
++    run: [/home/agent/.local/bin/claude-ollama]
+```
+
+The wrapper script:
+
+1. Sets `ANTHROPIC_BASE_URL` to `http://host.docker.internal:11434` (Ollama via Docker's host bridge)
+2. Uses a dummy `ANTHROPIC_AUTH_TOKEN` — Ollama doesn't require real credentials
+3. Maps every Claude model alias (Opus, Sonnet, Haiku, sub-agent) to `$CLAUDE_OLLAMA_MODEL`
+4. Calls `exec claude "$@"` — the real Claude Code CLI takes over from there
+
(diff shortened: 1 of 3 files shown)


ericcurtin:
**Missing `settings.containerSettings.claude: true`** — `nanoclaw` (the only other kit using `docker/sandbox-templates:claude-code-docker`) sets this. Its absence may cause Claude Code's sign-in screen to appear or prevent proper credential wiring through the sandbox proxy, even if bypassed via `ANTHROPIC_AUTH_TOKEN`. This is a meaningful omission given the purpose of the kit.

**`ANTHROPIC_AUTH_TOKEN` vs `ANTHROPIC_API_KEY`** — The wrapper sets both `ANTHROPIC_AUTH_TOKEN="ollama"` and `ANTHROPIC_API_KEY=""`. Claude Code's Ollama integration reference (`ollama/cmd/launch/claude.go`) uses `ANTHROPIC_AUTH_TOKEN`. However, `ANTHROPIC_API_KEY=""` (explicit empty string) may cause issues with some Claude Code versions that check for an empty key as an error condition rather than falling through to `ANTHROPIC_AUTH_TOKEN`. This should be verified.

**`host.docker.internal` not universally available on Linux** — On Linux Docker hosts, `host.docker.internal` is only available when `--add-host=host.docker.internal:host-gateway` is passed, which is not standard in Docker Sandboxes. CI/sandbox environments running on Linux may silently fail to connect. A note in the README or a `host.docker.internal` resolution check in the wrapper would help.

**`sbx run` usage in README is wrong** — The command shown is:
   ```
   sbx run claude-ollama --kit "..." ~/my-project
   ```
   Every other kit in the repo uses `sbx run --kit "..." <agent-name>` (flag before positional). `code-server` (a mixin) uses `sbx run <agent-name> --kit "..."`. The agent-first form may not be standard for new agent kits. Minor but a documentation nit.

**No `memory:` block** — All other `kind: agent` kits in the repo (`amp`, `nanobot`, `nanoclaw`, `openclaw`, `pi`) include a `memory:` block documenting the sandbox environment. `claude-ollama` omits it. This is a convention, not a hard requirement, but inconsistent.

**No `description` on `initFiles` entry** — Other kits include a `description:` on `initFiles` items (e.g., `code-server`). Minor nit.

alakae:
Thanks @ericcurtin for the detailed review. Here's what was addressed:

**1. `settings.containerSettings.claude: true`** — Added, matching nanoclaw.
One small note: this field doesn't appear to be documented at https://docs.docker.com/ai/sandboxes/customize/kits/#spec-reference ; if I'm reading that page correctly. Happy to keep it since nanoclaw uses it, but flagging in case the docs are intentionally trimmed or the field has changed.

**2. `ANTHROPIC_API_KEY`** — Changed from `export ANTHROPIC_API_KEY=""` to `unset ANTHROPIC_API_KEY`. 

**3. `host.docker.internal` on Linux** — Added a README note. I only have access to a macOS host and verified the full flow end-to-end there (Docker Desktop provides `host.docker.internal` automatically). Without a Linux GPU host available to test against, I opted for a documentation note rather than shipping an untested connectivity check in the wrapper.

**4. `sbx run` command order** — Fixed, `--kit` flag now precedes the agent name. 

**5. `memory:` block** — Added, following the amp pattern.

**6. `description:` on `initFiles`** — Added.

ericcurtin (approved):

alakae:
Hi @ericcurtin, sorry for the noise! No code changes were made. I only added a GPG signature to comply with the required signed commits policy, which unfortunately stripped the DCO sign-off lines. I've now rebased with --signoff to fix that. Everything should be green again. Thanks for your patience!

ejia-docker:
Thanks! And Congrats on being external contributor #1! Let us know if you have any feedback.

PR #16, opened 2026-05-04
Add Trivy kit

marklechner:
## Summary
First security-tool kit in contrib. Adds a `kind: agent` kit for the [Trivy](https://trivy.dev/) vulnerability scanner.

## Origin
Authored as a demo for an LDx3 talk on "boundary-first agentic work" — running security scanners (the very class of tool that just got compromised in the TeamPCP campaign) inside Docker sbx with a tight egress allowlist and pinned install.

- **Install is version- and SHA256-pinned per arch** (not curl|sh) — direct response to the tag-rewrite vector TeamPCP used in March 2026
- **Allowlist deliberately excludes `mirror.gcr.io`** (Trivy's default primary DB source) and forces GHCR-only via `TRIVY_DB_REPOSITORY` env. Trims trust surface from {GitHub + GHCR + Google} to {GitHub + GHCR}.
- **No docker.sock mount** by default; image-scanning documented as opt-in via per-sandbox policy + `docker save | trivy image --input` (README has details).
- **Entrypoint is `[bash, -l]`** rather than `[trivy]` because Trivy is a one-shot CLI; bash gives users a working shell with `trivy` on PATH and the workspace as cwd.

## Test plan
- [x] \`sbx kit validate ./trivy/\` → VALID
- [x] \`sbx run --kit ./trivy/ trivy <ws>\` end-to-end on macOS (Apple Silicon → arm64 sandbox)
- [x] Real \`trivy fs\` scan against vulnerable Python fixture (Django 1.11, requests 2.6.0) — surfaces expected CVEs
- [x] Boundary verification: non-allowlisted hosts return HTTP 403 from proxy
- [ ] CI: \`go test -v -count=1 -timeout 10m ./...\` (TCK; runs on PR)

### DIFF
--- a/trivy/spec.yaml
+++ b/trivy/spec.yaml
@@ -18,20 +18,14 @@ network:
     # github.com 302-redirects to release-assets.githubusercontent.com for binary downloads.
     - github.com
     - release-assets.githubusercontent.com
-    # Vulnerability DB pull (runtime; trivy-db is published as an OCI artifact on GHCR).
-    # We deliberately exclude Trivy's *default* primary mirror (mirror.gcr.io / Google) and
-    # force Aqua's GHCR-hosted artifact instead — see TRIVY_DB_REPOSITORY in `environment`.
-    # This trims the kit's external trust surface from {GitHub + GHCR + Google} to {GitHub + GHCR}.
+    # Vulnerability DB pull at runtime. Trivy's default chain is mirror.gcr.io (primary)
+    # with ghcr.io as a fallback for the trivy-db OCI artifact; we declare both openly in
+    # the allowlist rather than relying on env-var overrides to redirect the source.
+    # The trust footprint is intentionally explicit at the policy layer; tightening this
+    # is part of the v2 path (hardened-distribution-channel install).
+    - mirror.gcr.io
     - ghcr.io
     - pkg-containers.githubusercontent.com
-
-environment:
-  variables:
-    # Skip Trivy's default mirror.gcr.io fallback chain; pin DB source to Aqua's GHCR artifact.
-    # Trivy 0.61+ honors this env var; older versions need the --db-repository CLI flag instead.
-    TRIVY_DB_REPOSITORY: "ghcr.io/aquasecurity/trivy-db"
-    # Same idea for the Java DB (used by `trivy fs` against jar/pom files).
-    TRIVY_JAVA_DB_REPOSITORY: "ghcr.io/aquasecurity/trivy-java-db"
 
 commands:
   install:
--- a/trivy/spec.yaml
+++ b/trivy/spec.yaml
@@ -1,0 +1,68 @@
+schemaVersion: "1"
+kind: agent
+name: trivy
+displayName: Trivy
+description: "Aqua's open source vulnerability scanner — sandboxed because security scanners shouldn't be able to compromise their hosts (TeamPCP, March 2026)."
+
+agent:
+  image: "docker/sandbox-templates:shell-docker"
+  aiFilename: AGENTS.md
+  persistence: persistent
+  # Drop into bash with trivy on PATH and the workspace as cwd. Run `trivy fs .` to scan.
+  entrypoint:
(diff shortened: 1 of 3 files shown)


trivy/spec.yaml
```diff
@@ -0,0 +1,68 @@
...
+    TRIVY_JAVA_DB_REPOSITORY: "ghcr.io/aquasecurity/trivy-java-db"
+
+commands:
+  install:
+    # Install Trivy from a pinned, digest-verified GitHub release.
+    # Why not the official `install.sh`: that pattern (curl|sh) is exactly what TeamPCP
+    # weaponized in March 2026 against tag-rewriteable references. We pin a known-good
+    # release by version AND SHA256, in the kit, in git. To bump: change TRIVY_VERSION
+    # plus the per-arch SHA256 below (sourced from the release's checksums.txt).
+    - command: |
```
cdupuis, line 37: This should come in via `apk add/apt-get install` from the DHI package repository. We already got `trivy` in there.

trivy/spec.yaml
```diff
@@ -0,0 +1,68 @@
...
+    # force Aqua's GHCR-hosted artifact instead — see TRIVY_DB_REPOSITORY in `environment`.
+    # This trims the kit's external trust surface from {GitHub + GHCR + Google} to {GitHub + GHCR}.
+    - ghcr.io
+    - pkg-containers.githubusercontent.com
+
+environment:
+  variables:
+    # Skip Trivy's default mirror.gcr.io fallback chain; pin DB source to Aqua's GHCR artifact.
+    # Trivy 0.61+ honors this env var; older versions need the --db-repository CLI flag instead.
+    TRIVY_DB_REPOSITORY: "ghcr.io/aquasecurity/trivy-db"
```
cdupuis: Why use those env vars? Isn't that a potential attack vector? Malicious payload can just overwrite those and point at an empty DB?

trivy/spec.yaml
marklechner: thanks for this - really good point. let me adjust

trivy/spec.yaml
marklechner: you're right. fixing

marklechner: fix in the new commit drops these env vars entirely and adds `mirror.gcr.io` (Trivy's default primary DB source) to `allowedDomains`. trade-off is one extra host in the allowlist, but every required egress is now visible at the policy layer rather than redirected via an env override that the agent itself could subvert.

tightening the trust footprint further (both the install channel and the DB source) is the v2 path for me: install via a hardened-distribution channel rather than a pinned GitHub release, and let the package's own provenance pin the DB. happy to iterate when the runtime-side integration for that lands cleanly.

trivy/spec.yaml
marklechner: follow-up -> i tried wiring v2 with `dhi.io/alpine-base` as the kit's image and `apk add trivy` from the DHI repo. Two things came up that don't look like kit-author workarounds:

1. **Proxy CA trust**: sbx's egress proxy MITMs HTTPS for inspection. The shell-docker base used by other kits has the proxy CA pre-trusted; the DHI Alpine base doesn't (and shouldn't, since the proxy is sbx-specific). Injecting the CA into the standard Alpine TLS trust paths at install time didn't cause apk-tools to honour it - looks like its libfetch/libtls pathway is independent of OpenSSL's `/etc/ssl/cert.pem` and `SSL_CERT_FILE`.

2. **APK auth via the proxy**: the kit format supports `serviceAuth` for header injection (the pi kit uses it for Anthropic), but no current kit wires DHI APK auth that way, and the shape of the auth header for `dhi.io/apk` isn't documented in the public spec yet.

Both feel like sbx-side capability gaps rather than something to land in a kit's `commands.install`. Going to keep v1 (GitHub-pinned + env-fix from L32) for now and switch to the hardened-distribution path when those pieces are in place. happy to retest the moment they are.

ejia-docker (approved):

PR #23, opened 2026-05-05
Add Open Interpreter agent kit

ealeyner:
## Summary

- Adds an `open-interpreter/` agent kit for [Open Interpreter](https://www.openinterpreter.com/) — a natural language interface that writes and runs code (Python, JS, Shell) to complete tasks
- `auto_run: true` enabled in the default profile — safe here because the sandbox provides OS-level isolation
- Supports both OpenAI (GPT-4o default) and Anthropic (Claude) via proxy-managed credentials; both are declared in `credentials.sources`
- `serviceDomains` is intentionally narrow (only the two LLM API hosts) — a wildcard would TLS-intercept OI's own code execution traffic, breaking downloads and package installs
- Default profile dropped via `files/` at `/home/agent/.config/open-interpreter/profiles/default.yaml`
- Ships `computer.import_computer_api: true` so OI gets its built-in file/browser/OS tool surface

## Network policy rationale

OI executes arbitrary user-directed code. The `allowedDomains` list covers OI's own operational needs (PyPI, npm, apt, GitHub for Open Procedures) without a catch-all wildcard. Tasks that reach unlisted hosts can be unblocked by stacking a domain-specific mixin kit.

## Test plan

- [x] `sbx kit validate ./open-interpreter/` — VALID
- [ ] `cd open-interpreter && go test -v -count=1 -timeout 10m ./...`
- [ ] `sbx run --kit ./open-interpreter/ open-interpreter` with `OPENAI_API_KEY` set — OI REPL launches, `auto_run` active, profile loaded

### DIFF
--- a/open-interpreter/README.md
+++ b/open-interpreter/README.md
@@ -91,12 +91,21 @@ mixin kit with the extra domains.
 
 | Component | How |
 | --- | --- |
-| `open-interpreter` | `pip install open-interpreter` at creation time |
+| `uv` | `curl -LsSf https://astral.sh/uv/install.sh \| sh` at creation time |
+| `open-interpreter` | `uv tool install --python 3.12 open-interpreter` at creation time |
 | Default profile | Dropped via `files/` at `/home/agent/.config/open-interpreter/profiles/default.yaml` |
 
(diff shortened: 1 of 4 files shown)


mdelapenya:
Closing as superseded by #20, thanks!

PR #24, opened 2026-05-05
Add aider kit

ealeyner:
## Summary

- Adds a `kind: agent` kit for [Aider](https://aider.chat/), an AI pair programming tool
- Installs Aider via uv with Python 3.12 at sandbox creation time
- Wires Anthropic, OpenAI, and Gemini auth through the SBX proxy (`serviceDomains` + `serviceAuth`)
- Pre-seeds `~/.aider.conf.yml` with sensible defaults (model alias, auto-commits, analytics off)

## Spec choices worth flagging

**`--python 3.12` on the install command**: The base image ships Python 3.13, but aider's `numpy` dependency resolves to a version that only has prebuilt wheels for Python ≤3.12. The base image has no C compiler, so `uv tool install aider-chat` fails trying to build numpy from source. Pinning `--python 3.12` causes uv to download a standalone CPython 3.12 runtime (~28 MB) from `releases.astral.sh` (already covered by `*.astral.sh` in `allowedDomains`) and install aider against it. Confirmed working end-to-end.

**`*.astral.sh` wildcard in `allowedDomains`**: Same rationale as the openhands kit (#22) — the uv install script at `astral.sh` redirects binary and Python runtime downloads to `releases.astral.sh`. Without the wildcard, those fetches get a 403 from the proxy. `serviceDomains` stays narrow (only the three LLM API endpoints) to avoid TLS interception on install-time CDN traffic.

**`--with pip` on the install command**: Aider uses pip internally for some operations. Without this flag, `uv tool install` omits pip from the isolated environment and aider fails at runtime.

**No `aiFilename`/`memory:`**: Aider doesn't have a single canonical "memory file" analogous to `CLAUDE.md` or `AGENTS.md`. Users pass project conventions via `--read <file>` or `.aider.conf.yml`. The README covers this.

## Test plan

- [x] `sbx kit validate ./aider/` — passes
- [x] `sbx run --kit ./aider/ aider` — sandbox created successfully
- [x] Manual install verified: `aider 0.86.2` installed and `--version` returns correctly
- [x] `sbx policy log` — no blocked domains during install; `*.astral.sh`, `pypi.org`, `files.pythonhosted.org` all resolve as `forward-bypass`
- [ ] TCK (`go test`) — CI will run this; requires Docker, not available in this environment

## Origin

Designed and tested locally against a Docker SBX sandbox. Followed the amp kit pattern for auth wiring.


aider/spec.yaml
```diff
@@ -0,0 +1,85 @@
...
+    AIDER_MODEL: "sonnet"
+    AIDER_AUTO_COMMITS: "true"
+    AIDER_ANALYTICS: "false"
+    AIDER_CHECK_UPDATE: "false"
+
+commands:
+  install:
+    - command: "curl -LsSf https://astral.sh/uv/install.sh | sh"
+      user: "1000"
+      description: Install uv
```
dvdksn: uv is already preinstalled in the `shell` template

aider/README.md
```diff
@@ -0,0 +1,168 @@
...
+launched from this kit:
+
+### Anthropic (default)
+
+```console
+$ sbx secret set-custom -g \
+    --host api.anthropic.com \
+    --env ANTHROPIC_API_KEY \
+    --placeholder "sk-ant-{rand}" \
+    --value "$ANTHROPIC_API_KEY"
```
dvdksn: do these need to be created with `set-custom`? can't the kit use the pre-existing `anthropic`, `google`, `openai` secret services?

aider/README.md
```diff
@@ -0,0 +1,168 @@
...
+    --env OPENAI_API_KEY \
+    --placeholder "sk-{rand}" \
+    --value "$OPENAI_API_KEY"
+```
+
+Then override the model at run time:
+
+```console
+$ sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=aider" \
+    aider -e AIDER_MODEL=gpt-4o
```
dvdksn: There is no `-e` flag on `sbx run`

aider/spec.yaml
mdelapenya: Good catch, thanks — confirmed, `uv` ships pre-installed in `shell-docker` at `/usr/local/bin/uv`, already on PATH. Dropped the redundant install step. Same copy-paste pattern was in `openhands` (#22) and `open-interpreter` (#20) too, fixed those as well.

aider/README.md
mdelapenya: Yep, you're right — fixed. The kit's credentials never set `apiKey.proxyManaged`, so the engine never populated a placeholder env var, and Aider (via LiteLLM) checks that env var is *present* before it'll even attempt a request — so `secret set-custom` was papering over a spec gap, not something actually required. Set `proxyManaged: true` on all three credentials (matching `pi`/`nanobot`/`crush`'s existing pattern) and rewrote the README: `sbx secret set <service>` (or just the first-run wizard) is enough now. Same fix landed on `openhands` (#22).

aider/README.md
mdelapenya: You're completely right, and thanks for catching this — I'd initially misverified it myself. Confirmed by reading the actual `sbx run` source: there's no `--env`/`-e` flag on it at all (that only exists on `sbx exec`, a different command). There's genuinely no supported way to override a kit's `environment.variables` at run time today. Fixed by using Aider's own native `--model` flag instead (`sbx run ... aider -- --model gpt-4o`), since `sbx run <agent> -- <args>` passes args straight through to the entrypoint. Same class of bug existed in `openhands` (#22), fixed there too — though OpenHands has no CLI flag equivalent, so that one needs editing the seeded settings file in-sandbox instead.

aider/README.md
```diff
@@ -0,0 +1,146 @@
+# aider
+
+A standalone agent kit (`kind: agent`) for [Aider](https://aider.chat/), an
```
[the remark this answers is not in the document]
mdelapenya: Fixed in c63107c — updated to `kind: sandbox`.

aider/files/home/.aider.conf.yml
```diff
@@ -0,0 +1,15 @@
+# Default Aider configuration for Docker SBX.
+# Environment variables (AIDER_*) from the kit spec take precedence over these.
+# Edit this file inside the sandbox or mount your own via sbx run --env-file.
```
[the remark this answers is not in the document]
mdelapenya: Fixed in c63107c — removed the incorrect `--env-file` reference; pointed at the per-project `.aider.conf.yml` mechanism instead, which is real.

aider/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Configuration
+
+A pre-seeded `~/.aider.conf.yml` sets sensible defaults (model alias, auto-commits,
+analytics off). To customise:
+
+- **Inside the sandbox**: edit `~/.aider.conf.yml` directly — changes persist across
+  restarts.
+- **Per-project**: add an `.aider.conf.yml` at the root of your workspace.
+- **Coding conventions**: add an `CONVENTIONS.md` or pass `--read <file>` at launch.
```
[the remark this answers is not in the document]
mdelapenya: Fixed in c63107c.

aider/README.md
dvdksn: Actually we're just about to release v0.39.0 which will introduce support for `sbx run -e`, so maybe `sbx run -e AIDER_MODEL=` will work?

kiview (approved):

PR #31, opened 2026-05-07
Add git-ssh-sign mixin kit

dvdksn:
Mixin that configures git to sign commits and tags using the SSH key
forwarded from the host's SSH agent.

The signing config (gpg.format, commit.gpgSign, user.signingKey, etc.)
is written to /etc/gitconfig at install time so git reads it at process
startup. The actual public key is written to the path referenced by
user.signingKey from a global pre-commit hook, which fires after the
SSH agent is available and just before git performs the signing
operation. The global hook chains to any repo-local pre-commit hook.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/git-ssh-sign/README.md
+++ b/git-ssh-sign/README.md
@@ -1,0 +1,82 @@
+# git-ssh-sign
+
+A mixin that configures git to sign commits and tags using the SSH key
+forwarded from your host's SSH agent. Works with any agent kit
+(`claude`, `codex`, `cursor`, etc.).
+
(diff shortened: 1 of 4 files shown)


ejia-docker (dismissed):

dvdksn:
@ejia-docker just pushed some doc fixes 🧹

kiview (approved):

ilopezluna (approved):

dvdksn:
I don't have write access to this repository so need someone else to merge :-)

PR #33, opened 2026-05-10
feat(nanoclaw): v2 kit with pre-baked sandbox image for fast boot

ealeyner:
## Summary

Rework the nanoclaw kit around a **pre-baked sandbox image** and the **kits v2 format**, so launching nanoclaw is trivially simple and fast. The previous kit cloned, `pnpm`-installed, `tsc`-built nanoclaw *and* built its agent container image at first sandbox boot (5–12 minutes). All of that now ships inside `docker.io/ealeyner/nanoclaw-sbx`, and the kit only applies policy.

**Measured first boot with sbx v0.32.0** (Apple Silicon): `sbx create` ~21s → inner images seeded from the embedded archive in 17s → service CLI socket up ~1s → OneCLI gateway running in 24s → attach lands directly in the setup wizard. Total: well under a minute.

### What's in the image (`nanoclaw/Dockerfile`, built by `scripts/build-image.sh`)

- `FROM docker/sandbox-templates:claude-code-docker` + Node 22 + pnpm 10
- nanoclaw checkout @ pinned ref (v2.1.11), deps installed, TypeScript built, upgrade-state stamped (the service's tripwire refuses bare checkouts)
- OneCLI CLI binary
- `/opt/nanoclaw/images.tar`: `nanoclaw-agent:latest`, `ghcr.io/onecli/onecli:1.23.0`, `postgres:18-alpine` — loaded into the inner Docker daemon on first boot by a lock-guarded idempotent script (kicked off in the background by a startup command)

### spec.yaml (schemaVersion 2)

- `kind: sandbox` + `sandbox:` block (v2 renames), custom image, entrypoint baked as a real script
- `network.publishedPorts`: 3000 (webhooks), 10254/10255 (OneCLI dashboard/gateway). Note: sbx v0.32.0 validates the field but doesn't auto-bind yet; README documents `sbx ports --publish` as interim.
- `CONTAINER_IMAGE=nanoclaw-agent:latest` env override so nanoclaw uses the pre-loaded agent image instead of its per-checkout slug-derived tag
- Auth wiring (serviceDomains/serviceAuth/oauth/credentials) unchanged from the previous kit
- `commands.install` is gone entirely; three small idempotent `commands.startup` entries remain (docker socket perms, background image seed, npm proxy config)

### Recipe doc

`docs/recipe-prebaked-image-kit.md` captures the repeatable process (modelled on shelajev/hermes-sbx-kits) for converting other kits to this pattern — image/kit split, inner-image pre-seeding, validation/test/publish steps, and the gotchas hit while building this one (`COPY --chmod` directory perms, `docker save` 0600 tars, install-state tripwires, CLI version skew).

## Test plan

- [x] `sbx kit validate ./nanoclaw/` — VALID (sbx v0.32.0)
- [x] TCK validation/policy/oauth subtests pass locally; the container subtest needs the image published to Docker Hub (pulls `sandbox.image`) — image published via the release workflow
- [x] End-to-end on macOS/Apple Silicon: build image → `sbx template load` → `sbx create --kit ./nanoclaw` (21s) → inner images seed (17s) → service socket up (1s) → OneCLI gateway via pre-seeded images (24s, dashboard HTTP 200 from host through published port) → `sbx run` attach lands in setup wizard
- [x] CI green: TCK (incl. container test against the published image) and DCO both pass


dvdksn:
Thanks — the direction here is right (environment pre-baked into an image, spec reduced to policy), and the boot-time win is real. But I'd like to not frame this as a recipe yet, because one piece of the shape needs reconsidering: the embedded inner-image tar.

A multi-GB `images.tar` baked into the sandbox image is invisible to everything we can normally point at a container image (policy, scanning, signing), stores every inner image twice per sandbox, and is what forces the per-arch CI builds. Since the spec already opens registry egress, digest-pinned pulls at first boot — in the background, right where `nanoclaw-load-images` runs today — would keep nearly all of the speedup without the opaque blob. That leaves `nanoclaw-agent`, which only exists as a local build; the better fix there is getting that image published (upstream ideally), not embedding it.

Bigger picture: agents that orchestrate containers all hit this same "inner images" problem (this kit, hermes, and the others mentioned in the recipe doc), and each is hand-rolling the same workaround. That's a product gap we should raise against sbx — something like kit-declared inner images seeded by the runtime — rather than a pattern to document and copy. I'd suggest reframing `docs/recipe-prebaked-image-kit.md` as notes on a transitional approach in the meantime.

Happy to help with the rework.

— authored and posted by Claude

ealeyner:
**Commit ef66a4a** — two files changed.

**Root cause:** nanoclaw never reads \`ANTHROPIC_API_KEY\` from the process environment. Credentials must live in the OneCLI vault (\`onecli secrets create --type anthropic\`). On first boot the vault is empty, so \`runAuthStep()\` in \`setup/auto.ts\` always falls through to the interactive auth picker — even though the sandbox has already injected the key.

**Validated:** \`--type anthropic\` is nanoclaw's own native secret type (used by \`createAnthropicSecret()\` and \`runPasteAuth()\`). It tells the OneCLI gateway to inject the \`x-api-key\` header — correct for \`api.anthropic.com\`. The alternative \`type=generic + Authorization: Bearer\` path (\`runCustomEndpointAuth()\`) would 401 on every container agent request.

---

**\`scripts/start-nanoclaw.sh\`**
- Runs the OneCLI gateway installer after \`seed-images\` (images are pre-cached, so the pull is a no-op; installer just brings the compose stack up).
- Polls \`/api/health\` until healthy before proceeding.
- Pre-registers \`ANTHROPIC_API_KEY\` as \`type=anthropic\` so \`anthropicSecretExists()\` returns true and the interactive auth picker is skipped.
- Passes \`NANOCLAW_ONECLI_API_HOST=http://127.0.0.1:10254\` on the \`exec\` so the onecli setup step takes the non-interactive \`--remote-url\` path (no fresh/reuse prompt, no second installer run, no conflict with the already-running gateway).
- \`npm run\` → \`pnpm run\` to match vanilla nanoclaw.
- Moves \`export PATH\` before the gateway bootstrap block (needed for the \`onecli\` CLI binary).

**\`spec.yaml\`**
- Adds missing channel API domains: \`api.telegram.org\`, \`discord.com\`, \`gateway.discord.gg\`, \`api.slack.com\`, \`slack.com\`. Without these, token validation during \`/add-telegram\` etc. and runtime Discord callbacks fail.
- Extends the proxy startup command to also run \`pnpm config set proxy/https-proxy\` explicitly (belt-and-suspenders alongside the \`npm config set\` that writes \`~/.npmrc\`).

**Known remaining gaps (not in this commit):**
- GAP 2: nanoclaw service needs a restart after \`/add-channel\` installs a new adapter (the process doesn't auto-reload).
- GAP 3: \`verify\` step checks \`launchd\`/\`systemd\` service status; the sandbox runs the process directly, so that check may report \`service=not_found\`.

mdelapenya:
@ealeyner I think we must close this PR, as the Nanoclaw folks are hosting their Dockerfile at the moment, publishing it under their own Docker organisation. Besides, v2/work was already merged, so this PR got outdated.

Feel free to reopen it if you consider it otherwise 🙏 

Thanks again for your hard work!

PR #39, opened 2026-05-14
feat(opencode-model-runner): use opencode-models-discovery plugin

chris-crone:
Switch from a hardcoded qwen3 model config to the opencode-models-discovery plugin, which queries Docker Model Runner's models endpoint at startup. Any model pulled with `docker model pull` is automatically available in OpenCode's model picker without manual config edits.

Screenshot 2026-05-14 at 09 28 32


doringeman (approved):
Thanks!

Image

mdelapenya (approved):

PR #40, opened 2026-05-15
feat(crush): add crush agent kit

mdelapenya:
## Summary

Adds a standalone agent kit for [Crush](https://github.com/charmbracelet/crush), Charm's multi-provider AI coding agent.

- `kind: agent` on `docker/sandbox-templates:shell-docker`, entrypoint `crush --yolo`
- Installs via the official Charm apt repository (`repo.charm.sh`)
- 16 proxy-managed provider env vars and 15 credential sources covering Anthropic, OpenAI, Azure OpenAI, Google Gemini, Mistral, Groq, Cerebras, OpenRouter, Hugging Face, io.net, MiniMax, Synthetic, Vercel, Z.ai, and AWS Bedrock
- Matching `serviceDomains` + `serviceAuth` so the proxy injects the right header per provider; `allowedDomains` covers the install repo and every provider API host

## Test plan

- [x] `go test -short ./crush/...` — pure-logic TCK passes
- [x] `go test ./crush/...` — full TCK passes, including live install of `crush` from `repo.charm.sh`
- [x] `go vet` and `gofmt` clean
- [x] Manually verified end-to-end on a temporary sandbox

### DIFF
--- a/crush/spec.yaml
+++ b/crush/spec.yaml
@@ -1,0 +1,178 @@
+schemaVersion: "1"
+kind: agent
+name: crush
+displayName: Crush
+description: Multi-provider AI coding agent from Charm, with proxy-mediated auth for 15 model providers.
+
+agent:
+  image: "docker/sandbox-templates:shell-docker"
+  aiFilename: AGENTS.md
+  entrypoint:
+    run: [crush]
+    args: ["--yolo"]
+
+environment:
+  proxyManaged:
+    - ANTHROPIC_API_KEY
+    - AWS_ACCESS_KEY_ID
+    - AZURE_OPENAI_API_KEY
+    - CEREBRAS_API_KEY
+    - GEMINI_API_KEY
+    - GOOGLE_GENERATIVE_AI_API_KEY
+    - GROQ_API_KEY
+    - HF_TOKEN
+    - IONET_API_KEY
+    - MINIMAX_API_KEY
+    - MISTRAL_API_KEY
+    - OPENAI_API_KEY
+    - OPENROUTER_API_KEY
+    - SYNTHETIC_API_KEY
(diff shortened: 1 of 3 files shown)


crush/spec.yaml
```diff
@@ -0,0 +1,178 @@
...
+    api.minimax.chat: minimax
+    api.synthetic.com: synthetic
+    api.zai.com: zai
+    v0.dev: vercel
+    "bedrock-runtime.*.amazonaws.com": aws
+    "bedrock.*.amazonaws.com": aws
+  serviceAuth:
+    anthropic:
+      headerName: x-api-key
+      valueFormat: "%s"
```
dvdksn, line 88: do we need to redefine the serviceAuth for specs if the service already exists as a built-in? eg anthropic, github, openai

ilopezluna (approved):

crush/spec.yaml
kiview: Technically we don't. Technically, I am also not sure, how it merges with the existing service.
Conceptually, having kits more self-contained sounds somewhat good.

mdelapenya: Mmm in this case, we use shell as base. You mean in the case a mixin is installed on an agent?

dvdksn: I meant for agents too actually. `anthropic`, `google`, etc are pre-defined "services":

```
[david:~/s/g/d/docs] main+ ± sbx secret set --help
Create or update a secret for a service.

Available services: anthropic, aws, cursor, droid, github, google, groq, mistral, nebius, openai, xai
```

So it feels like these are somehow baked in and available by default, even for custom agents. But I'm honestly not sure what "Available servics" really means.

PR #42, opened 2026-05-20
spec: add MCPClient schema

slimslenderslacks:
## Summary

Adds `MCPClientSpec`, `MCPClientFile`, and `MCPClientCommand` to the
`spec/` package so agent kits can declare how the consuming engine
should configure the agent to talk to a sandbox's aggregate MCP
gateway.

The schema is intentionally data-only — Go `text/template` rendering
and the template-error type stay engine-side because they depend on
engine-only context (gateway URL, sentinel token, runtime / workspace
dirs). Kits describe the *what* (file paths, file content, exec argv,
all as template strings); the engine decides the *when, where, and as
whom*.

### What lands here

- `spec/mcp_client.go` — `MCPClientSpec`, `MCPClientFile`, `MCPClientCommand` with `json` + `yaml` tags.
- `spec/types.go` — `MCPClient *MCPClientSpec` on `Artifact` and the on-disk `specFile`.
- `spec/artifact.go` — round-trip `MCPClient` through `parseArtifact`.
- `spec/validate.go` — `ValidateMCPClientSpec` (rejects empty paths/contents, empty argv, out-of-range modes) and wires it into `ValidateArtifact`.

### Why now

The `docker/sandboxes` engine has a `mcp-kit-hooks` branch landing
soon that consumes this schema. The engine ships an executor-agnostic
`applyMCPClientHook` that:

1. Looks up the kit's `MCPClientSpec` when the gateway starts.
2. Renders each `Files[i].Path` / `Files[i].Content` and each
   `Commands[i].Argv[j]` as a Go template with a fixed context
   (`GatewayURL`, `SentinelToken`, `RuntimeID`, `WorkspaceDir`,
   `HomeDir`).
3. Writes the rendered files inside the running container, then
   execs the rendered commands.

A claude kit using it looks like:

```yaml
mcpClient:
  commands:
    - argv:
        - sh
        - -c
        - |
          set -e
          cd "{{.WorkspaceDir}}"
          claude mcp add mcp-gateway "{{.GatewayURL}}" --transport http --header "Authorization: Bearer {{.SentinelToken}}"
```

### Compatibility

Purely additive. `MCPClient` is a pointer; existing kits decode and
validate unchanged. Engines that don't know about the field ignore
it.

## Test plan

- [x] `go build ./spec/` passes
- [x] `go test ./spec/` passes (existing suite unaffected)
- [ ] Reviewer: confirm json/yaml field names match the
      sandboxes-side consumer (`docker/sandboxes` branch
      `mcp-kit-hooks`).
- [ ] Reviewer: confirm `ValidateMCPClientSpec` bounds (mode range,
      whitespace-only path rejection) feel right.

### DIFF
--- a/spec/artifact.go
+++ b/spec/artifact.go
@@ -112,6 +112,7 @@ func parseArtifact(readFile func(string) ([]byte, error)) (*Artifact, error) {
 		Settings:    spec.Settings,
 		Commands:    spec.Commands,
 		OAuth:       spec.OAuth,
+		MCPClient:   spec.MCPClient,
 		Memory:      spec.Memory,
 	}, nil
 }
--- a/spec/mcp_client.go
+++ b/spec/mcp_client.go
@@ -1,0 +1,47 @@
+package spec
+
+// MCPClientSpec declares how a kit configures its agent to talk to the
+// sandbox's aggregate MCP gateway. The consuming engine reads this spec
+// when a runtime starts the gateway and applies the declared files and
+// commands inside the container.
+//
+// The spec is intentionally declarative: paths, contents, and command
+// argv elements are Go text/template strings rendered against a context
+// supplied by the engine (gateway URL, sentinel token, runtime/workspace
+// dirs). The spec library only defines the data shape; template parsing
+// and rendering live in the consumer.
+type MCPClientSpec struct {
+	// Files to write inside the container after the gateway is up. Each
+	// file's Path and Content are rendered as Go templates by the
+	// consumer.
+	Files []MCPClientFile `json:"files,omitempty" yaml:"files,omitempty"`
+
+	// Commands to exec inside the container after the files are written.
+	// Use sparingly — most agents need only a config file write.
+	Commands []MCPClientCommand `json:"commands,omitempty" yaml:"commands,omitempty"`
+}
+
+// MCPClientFile describes one file the kit needs the engine to write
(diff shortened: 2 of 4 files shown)


bendebury:
A couple of concerns:

1. The kit spec already has `commands.initFiles` for file creation and `commands.startup` for setup commands. The dedicated MCP block seems to reproduce this functionality in an mcp-only block. I am concerned both about the precedent this sets and the unclear ways that this block will interact with the previously mentioned blocks. Would we then add dedicated blocks for future protocols that follow the "I need special files and commands" pattern?

2. Why not do this via mixins?

slimslenderslacks:
Design changed: gateway-first lifecycle in docker/sandboxes (commit 6bca3fd6 on the mcp-kit-hooks branch there) lets agent kits register the MCP gateway via existing `commands.startup` + container env vars (`MCP_GATEWAY_URL`, `MCP_SENTINEL_TOKEN_NAME`), with no schema additions to spec needed.

The `MCPClient*` types this PR introduced are no longer consumed by the engine — see the parallel cleanup at https://github.com/docker/sandboxes (branch `mcp-kit-hooks`, commit `mcp: delete mcpClient schema and machinery`).

Closing without merging. The commit on this branch (`0c43710`) stays available if the design ever rotates back; we'd cherry-pick from there.

PR #46, opened 2026-05-22
feat(network): add publishedPorts to NetworkPolicy

ndeloof:
## Summary

Adds a declarative `network.publishedPorts` field to `NetworkPolicy` so kit authors can tell the sandbox runtime to publish in-container ports on the host when the sandbox starts.

```yaml
network:
  publishedPorts:
    - container: 9418
    - container: 8080
      protocol: tcp
      name: web
    - container: 53
      protocol: udp
      name: dns
```

## Why

Today's UX:
- `code-server` kit's README ships an explicit `sbx ports --publish 8080` step the user must run after `sbx run`
- The same gap blocks moving git-daemon (used by `--clone` in docker/sandboxes) into a kit
- Future kits (Jupyter, dev servers, …) all hit the same wall

A declarative field lets kits express the port surface they want exposed, and lets the engine wire it once at create time.

## Schema

| Field | Type | Required | Notes |
|-------|------|----------|-------|
| `container` | int | yes | 1..65535 |
| `protocol` | string | no | `tcp` (default) or `udp` |
| `name` | string | no | informational label for tools like `sbx ports` |

Host port is **always ephemeral** (127.0.0.1:`<assigned>`). Users wanting a pinned host port keep using `sbx ports --publish 8000:8080` — that surface is unchanged. The split is deliberate: a kit can't pick a host port because two kits requesting the same one would collide on the user's machine.

## Validation

- `container` outside `1..65535` → error per entry index
- `protocol` other than `""`, `"tcp"`, `"udp"` → error per entry index
- Empty `protocol` is accepted; consumers default it to `"tcp"` at use-site (keeps the struct ergonomic for YAML omit)

## Consumer wiring

Lives in `github.com/docker/sandboxes` — separate follow-up PR. After this PR is tagged, the engine will:
1. Bump `sbx-kits-contrib` in `go.mod`
2. Merge `PublishedPorts` across composed kits (next to existing `AllowedDomains` / `DeniedDomains` merging)
3. Call `rtMgr.PublishPort` for each entry after sandbox creation

Cleanup is free — sandboxd's `cleanupPublishedPorts` runs on every `DeleteRuntime`.

## Test plan

- [x] `go test ./spec/` — passes
- [x] Validation: container range, protocol whitelist
- [x] YAML round-trip: minimal (container only) + full (container + protocol + name)

## Release

Please tag once merged so docker/sandboxes can pin the new version.


mdelapenya (dismissed):
LGTM, although for completeness, can you add the new fields to the testdata kits?

ndeloof:
Thanks @mdelapenya! Added `network.publishedPorts` to both testdata kits in 7689626:
- `spec/testdata/sample-mixin/spec.yaml` — full coverage: minimal (container only), tcp+name, and udp+name forms
- `spec/testdata/sample-agent/spec.yaml` — a single tcp `web` port

`go test ./spec/... ./tck/...` and `go vet` both pass, so the TCK self-tests now load and validate the field end-to-end.

dvdksn (approved):

PR #48, opened 2026-05-22
add github-ssh kit for SSH host key bootstrapping

dvdksn:
## Summary

Adds a new `github-ssh` kit that pre-populates `~/.ssh/known_hosts` with GitHub's host keys at install time. This allows SSH operations to GitHub to work in sandboxes without interactive host verification prompts, which fail because there is no TTY.

The kit is intentionally separate from `git-ssh-sign` — SSH host key bootstrapping is specific to GitHub, while commit signing is host-agnostic.

**Host prerequisites** (documented in the kit's memory): SSH agent forwarding must be enabled when creating the sandbox, and the key in the agent must be registered with the user's GitHub account.

Generated by Claude Code


dvdksn:
I thought this worked but on second look it seems something is missing

ejia-docker (approved):

PR #50, opened 2026-05-23
Add packages-through-sfw kit

VisibleSampling:
## Summary

  Adds a `socket-firewall` mixin kit that installs Socket Firewall Free (`sfw`) and routes npm/PyPI package-manager commands through it when
  resolved through PATH.

  - Installs pinned Socket Firewall Free v1.10.0 with per-arch SHA256 checks.
  - Uses PATH shims for `npm`, `pip`, `pip3`, and `python3 -m pip` so non-interactive agent tool calls are covered.
  - Direct paths such as `/usr/bin/npm` and venv paths such as `venv/bin/python -m pip` intentionally bypass Socket Firewall; SBX egress still
  applies.
  - Pip wrappers unset Python CA override variables so `sfw pip` works under the sandbox proxy.

  ## Test plan

  - Fast TCK/spec validation passed.
  - Manual `sbx run --kit ./socket-firewall codex`
  - Verified `npm`, `pip`, `pip3`, and `python3 -m pip` route through Socket Firewall.
  - Full TCK container test was not run locally because rootless Docker was unavailable.

### DIFF
--- a/packages-through-sfw/README.md
+++ b/packages-through-sfw/README.md
@@ -1,0 +1,89 @@
+# Packages through SFW
+
+A mixin that installs [Socket Firewall Free](https://socket.dev/) (`sfw`)
+and routes `npm`, `pip`, `pip3`, and `python3 -m pip` package-manager
+commands through it inside a Docker Sandbox when they are resolved through
+`PATH`.
+
+Socket Firewall Free runs without Socket API keys or configuration files.
+This kit targets public npm and PyPI package installs. Private registries,
+custom registries, and organization policy features require Socket Firewall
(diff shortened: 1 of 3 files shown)


VisibleSampling:
Fixed signature blocker

ejia-docker (dismissed):

ejia-docker:
Thinking about this change a bit, is it a little misnamed? It installs SFW, but also does the routing of package managers through it, which might be surprising to others. 

Perhaps packages-through-sfw or something?

VisibleSampling:
I like it, updated.

ejia-docker (approved):

PR #51, opened 2026-05-24
Add hermes-agent kit

MonojitBanerjee:
**Summary**
  - Adds a kind: agent kit for Hermes Agent by Nous Research — a self-improving AI agent that creates skills from experience, supports 200+ models, and includes a built-in
  cron scheduler and multi-platform gateway.
  - Installs via the official install.sh in a detached background session; a polling entrypoint wrapper waits for the install flag before exec-ing hermes. Install logs land
  at ~/hermes-install.log.
  - Supports three providers out of the box: Anthropic, OpenAI, and OpenRouter — each wired through the sandbox proxy.

  **Spec choices worth flagging**

  - Three credentials sources + proxyManaged — Hermes is multi-provider by design, so ANTHROPIC_API_KEY, OPENAI_API_KEY, and OPENROUTER_API_KEY are all declared. The user
  exports whichever key they have; sbx reads it via credentials.sources and the proxy injects it on outbound calls to the matching serviceDomain. Only the relevant key needs
   to be present.
  - Background install with setsid — The official installer takes ~3–5 minutes (uv + Python 3.11 + pip install -e ".[all]"). Running it in a detached session keeps sandbox
  creation non-blocking, matching the pattern used by nanoclaw and openclaw.
  - shell-docker base image — Hermes supports a Docker environment backend, so Docker-in-Docker availability in shell-docker is intentional.
  - astral.sh in allowedDomains — needed for the uv installer; the binary itself is fetched from github.com/objects.githubusercontent.com which are also listed.

  **Test plan**

  - sbx kit validate ./hermes-agent/ — passes
  - go test -v -count=1 -timeout 10m ./... — all 12 TCK subtests pass
  - sbx run --kit ./hermes-agent/ hermes-agent — sandbox creates, installs, and launches hermes successfully

  **Origin**

  Hermes Agent is an open-source project by Nous Research, MIT-licensed, hosted at https://github.com/NousResearch/hermes-agent.


mdelapenya:
HI @MonojitBanerjee thanks for the PR! Can you update it removing the TCK tests? We simplified the testing story so kits do not need to include the TCK, as the automation in place will do that instead.

On other topic, I'm really interested in your use case for Hermes, sbx and kits, could you share a bit if possible? 🙏

MonojitBanerjee:
Thanks @mdelapenya . I have updated the PR as per your request. 
For question related to usage- hermes is an AI agent for automation, coding etc.  With this change, it can run in sandbox  easily and can be deployed securely. Right now running it  locally is mostly fine for individual use case but via sanbox will provide isolation which is good for large scale deployment.

mdelapenya:
I noticed all these domains are blocked when I run the sandbox with a deny-all base policy. Our recommendation is kits must provide all their allowed domains.

Screenshot 2026-05-29 at 17 46 35

The docker.com and ubuntu domains are usually added, see codex-app-server kit, although I'm interested in the 140.82.113.4 and duckduckgo.com ones.

Do you know why they are needed?

MonojitBanerjee:
Thanks @mdelapenya for your review. I have added an inline comment about why duckduckgo.com is needed(install.sh connectivity probe alongside pypi.org comes from official hermes install script)

For 140.82.113.4 - its not needed explicitly. 

```
Bash(dig -x 140.82.113.4 +short 2>/dev/null || nslookup 140.82.113.4 2>/dev/null | head -10)
  ⎿  lb-140-82-113-4-iad.github.com.
```

● That's GitHub's infrastructure (lb-140-82-113-4-iad.github.com — a GitHub load balancer in the IAD/Virginia region). It's not hermes itself making that call — it's the install script during sandbox creation:

`  curl -fsSL https://raw.githubusercontent.com/NousResearch/hermes-agent/main/scripts/install.sh | bash
`

 That curl resolves raw.githubusercontent.com → GitHub's IPs, then the installer clones the repo from github.com and downloads the uv binary from GitHub releases — all of which resolve to 140.82.113.x addresses.

Can you confirm why Sandbox network policy is not allowing it? Should I add *.github.com too? I already added github.com

Also from my local- I see this (and  not that IP).  

image

mdelapenya (approved):
LGTM, thanks for this contribution!

I do not have an OpenRouter key, but other than that, I can see the Hermes TUI.

It could be of interest if we add `openrouter` as a built-in secret provider (like anthropic, openai, ...), so the agent transparently reads the secret from the secrets manager that is built-in in sbx. Will ping you back here if we add it.

PR #53, opened 2026-05-26
Add Junie agent kit

preigile:


mdelapenya:
Hi @preigile thanks for your contribution! It LGTM although the commits need to be signed 🙏 https://github.com/docker/sbx-kits-contrib/blob/main/CONTRIBUTING.md#sign-off-and-signing

preigile:
Thank you @mdelapenya for your review! I signed my commit.

junie/junie_tck_test.go
```diff
@@ -0,0 +1,14 @@
+package junie_test
```
mdelapenya: suggestion: could you remove this test file?

We have updated all the kits so that the TCK does not need to add any test file here unless specific, non-in-the-tck tests are needed

kiview (dismissed):

junie/junie_tck_test.go
preigile: Yes, sure! Done

junie/spec.yaml
```diff
@@ -0,0 +1,80 @@
+schemaVersion: "1"
+kind: agent
+name: junie
+displayName: Junie
+description: The AI coding agent by JetBrains.
+
+agent:
+  image: "docker/sandbox-templates:shell-docker"
+  aiFilename: .junie/AGENTS.md # Junie also supports AGENTS.md
```
dvdksn: @mdelapenya IIUC, this means that the sbx-contextual agent instructions ("you are running in a sandbox, here's how your environment works") would land in `<workspace>/../.junie/AGENTS.md`.

@preigile does Junie traverse the parent hierarchy and look for `.junie` directories at each level above CWD, or does it only look for `.junie/AGENTS.md` in cwd? Other agents like Claude/Codex search upward so a `./CLAUDE.md` and `../CLAUDE.md` both compound to an agent's context.

preigile: @dvdksn Junie doesn't traverse the parent directory hierarchy. It looks for guidelines in the following order:
1. .junie/AGENTS.md file in the project root.
2. AGENTS.md file in the project root.

mdelapenya: I'd suggest leaving it as `AGENTS.md` only. Because Junie is a coding agent, sbx will create that file one folder above the copied workspace, and it does not support other paths for that (like `.junie/AGENTS.md`). So the `aiFilename` is not a description of what the agent supports, but public API that decides where sbx will write down some agent context, like project info we auto-detect (i.e. programming language for tools)

Sorry if this is not that clear 🙏

preigile: No problem. Done

mdelapenya (approved):

mdelapenya:
Merged, thanks for adding Junie! 🙇

PR #59, opened 2026-05-29
spec: redesign volumes (add MountSpec.Type, drop Manifest.Tmpfs) — v2 Phase 2

bendebury:
## Summary

Phase 2 of the v2 unified kit-spec migration. Single breaking change in this PR:

- Adds `MountSpec.Type` (valid values: `""` for the default block-backed mount, `"tmpfs"` for a RAM-backed mount).
- Drops `Manifest.Tmpfs`. Tmpfs entries now live in `Manifest.Volumes`, distinguished by `Type`.
- Adds `Manifest.TmpfsVolumes()` helper for callers that previously read the separate slice. The `tck` package uses it.
- `ValidateTmpfs` is gone; `ValidateVolumes` covers all entries and gains a Type-enum check.

**Breaking change.** v1 kits with a top-level `tmpfs:` block strict-decode-reject at load time with:

> `field tmpfs not found in type spec.specFile`

Kit authors migrate by moving each `tmpfs:` entry into `volumes:` with `type: tmpfs`:

```yaml
# Before (v1)
tmpfs:
  - path: /tmp/scratch
    size: 512m

# After (v2)
volumes:
  - path: /tmp/scratch
    type: tmpfs
    size: 512m
```

The sandboxes-side consumer changes (and the corresponding pseudo-version pin) ship in the matching docker/sandboxes PR, which is stacked on the in-flight Phase 1 sandboxes PR (#3177 on docker/sandboxes).

## Test plan

- [x] `go test ./spec ./tck` green
- [x] Three new spec tests: `TestVolumesType_TmpfsAccepted`, `TestV1TmpfsBlock_StrictRejected`, `TestValidateArtifact_MountSpecType`
- [x] No built-in or contrib kits used the v1 `tmpfs:` block (zero kit spec.yaml migrations needed in-tree).

## Refs

- Design doc: docker/sandboxes `docs/specs/2026-05-27-unified-kit-spec-v2.md` (Phase 2)
- Phase 2 plan: docker/sandboxes `docs/specs/2026-05-28-unified-kit-spec-v2-phase2-plan.md`

### DIFF
--- a/spec/types.go
+++ b/spec/types.go
@@ -104,16 +104,16 @@ type Manifest struct {
 	Volumes []MountSpec `json:"volumes,omitempty" yaml:"volumes,omitempty"`
 }
 
-// TmpfsVolumes returns the subset of m.Volumes whose Type is "tmpfs".
-// Convenience for call-sites that previously read the separate
-// Manifest.Tmpfs field.
+// TmpfsVolumes returns the subset of m.Volumes whose Type is
+// MountTypeTmpfs. Convenience for call-sites that previously read the
+// separate Manifest.Tmpfs field.
 func (m *Manifest) TmpfsVolumes() []MountSpec {
 	if len(m.Volumes) == 0 {
 		return nil
 	}
 	out := make([]MountSpec, 0, len(m.Volumes))
 	for _, v := range m.Volumes {
-		if v.Type == "tmpfs" {
+		if v.Type == MountTypeTmpfs {
 			out = append(out, v)
 		}
 	}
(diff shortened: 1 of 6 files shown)


spec/types.go
```diff
@@ -112,14 +129,18 @@ type Security struct {
...
+// the backing storage: omit or set "" for the default block-backed
+// volume; set "tmpfs" for a RAM-backed mount.
 type MountSpec struct {
 	// Path is the absolute mount path in the container.
 	Path string `json:"path" yaml:"path"`
 
+	// Type selects the backing storage. Omit (or set "") for the default
+	// block-backed volume; set "tmpfs" for a RAM-backed mount. Added in
+	// schemaVersion "2" to replace the separate top-level `tmpfs:` block.
+	Type string `json:"type,omitempty" yaml:"type,omitempty"`
```
mdelapenya: suggestion: I think we could expose the Type values as constants, even with a MountType type that downstream consumer can use to write tooling around kit generation without the need of simple (and error prone) string.

mdelapenya (dismissed):
LGTM! I added one comment regarding exposing the volume types, fine to be done in a follow up if you like it

spec/types.go
bendebury: Done in `b6d8d29` (followed by `1c96a46` after the merge from main). Added a `MountType` named type with constants `MountTypeBlock` (`""`) and `MountTypeTmpfs` (`"tmpfs"`). YAML/JSON encoding is unchanged — `type MountType string` marshals as a string and untyped string literals in existing call sites still compare against the new type without conversion. The validator and the `Manifest.TmpfsVolumes()` helper use the constants now.

The corresponding sandboxes-side update (re-export the constants from `sandboxlib/kit/types.go` and switch `agent.go`'s volumes partition to `MountTypeTmpfs`) is in docker/sandboxes#3209, commit `668f475e`. Thanks for the suggestion — much nicer for tooling-side consumers.

mdelapenya (approved):

PR #63, opened 2026-05-29
scripts: add migrate-v1-to-v2 for v2 spec migration

bendebury:
## Summary

Standalone Go script kit authors run to convert their `spec.yaml` from schemaVersion 1 to schemaVersion 2. Lands ahead of the rest of the v2 migration so kit authors have a working tool the moment the breaking changes start shipping.

The script is intentionally **incremental**. Today's scope covers Phase 1 cosmetic renames; later phases (volumes redesign, credentials redesign, settings lift) extend the same `applyPhase1Transforms` pattern as their PRs land. See `scripts/README.md` for the current scope and the [v2 spec doc](https://github.com/docker/sandboxes/blob/main/docs/specs/2026-05-27-unified-kit-spec-v2.md) on docker/sandboxes for the migration roadmap.

## Usage

```bash
go run scripts/migrate-v1-to-v2.go <path-to-kit-directory>
```

Writes `<path>/spec.yaml` in place and leaves `<path>/spec.yaml.bak` as a backup. Running on an already-v2 spec is a no-op and does not create a `.bak`.

## Transforms (Phase 1)

| v1 spelling | v2 spelling |
|---|---|
| `kind: agent` | `kind: sandbox` |
| `agent:` block | `sandbox:` block |
| `memory:` field | `agentContext:` field |

## Design

- **Standalone module-free** — `package main` with only stdlib imports, so kit authors can `go run` against any checkout without pulling sbx-kits-contrib's deps.
- **Regex line replacement** — preserves comments, blank lines, and block-scalar formatting. A YAML re-marshal would have reformatted the entire file; the regex approach keeps the diff to just the renamed fields.
- **Refuses to clobber an existing `.bak`** — if a previous migration left one, the script errors rather than silently overwriting the backup.
- **Anchored regexes** — top-level YAML keys only. Nested `agent:` (unlikely but possible) is left alone.

## Test plan

- [x] Golden-file tests under `scripts/testdata/` — v1 input + v2 expected pairs compared byte-for-byte
- [x] End-to-end tests via the `migrate()` function against temp-dir copies of the fixtures
- [x] No-op verification on already-v2 specs
- [x] Refuse-to-clobber-`.bak` verification
- [x] Missing-spec verification
- [x] `go test ./scripts/...` green

## Docs

- New `scripts/README.md` — canonical usage reference, scope-as-it-grows, fixture conventions
- `CONTRIBUTING.md` adds a "Migrating an existing kit to v2" section pointing at the script

### DIFF
--- a/scripts/migrate_v1_to_v2_test.go
+++ b/scripts/migrate_v1_to_v2_test.go
@@ -1,0 +1,191 @@
+package main
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestApplyPhase1Transforms_FullShape exercises a v1 spec.yaml that uses
+// every Phase 1 renamed field at once and confirms the output matches
+// the v2-expected golden file byte-for-byte. Comments, blank lines, and
+// block-scalar formatting must survive.
+func TestApplyPhase1Transforms_FullShape(t *testing.T) {
+	input, err := os.ReadFile(filepath.Join("testdata", "v1-full", "spec.yaml"))
+	if err != nil {
+		t.Fatalf("read input: %v", err)
+	}
+	expected, err := os.ReadFile(filepath.Join("testdata", "v2-expected", "spec.yaml"))
+	if err != nil {
+		t.Fatalf("read expected: %v", err)
+	}
+
+	got, changes := applyPhase1Transforms(string(input))
+
+	if got != string(expected) {
+		t.Errorf("output mismatch.\nGOT:\n%s\n\nEXPECTED:\n%s", got, expected)
+	}
+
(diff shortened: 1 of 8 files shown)


scripts/migrate_v1_to_v2_test.go
```diff
@@ -0,0 +1,191 @@
+package main
```
mdelapenya, line 1: bug: these tests are not connected to the CI, so we need to add the `scripts` dir to the detect-changes CI step. Else, the CI won't run the tests in the dir.

bendebury: Should be done now!

mdelapenya (approved):

PR #64, opened 2026-06-02
Phase 3: credentials[] + caps.network + bindings file format + parsers

bendebury:
## Summary

Phase 3 of the unified kit-spec v2 migration restructures four overlapping v1 surfaces (`credentials.sources` map, `network.serviceAuth`, `network.serviceDomains`, `environment.proxyManaged`, standalone `oauth:` block, `network.allowedDomains`/`deniedDomains`) into:

- A unified `credentials[]` list per service (apiKey + oauth sub-blocks; inject domains/headers)
- A `caps.network.allow/deny` block for the egress allow/deny lists
- A separate user-side bindings file (`~/.config/sbx/credentials.yaml`) declaring where credentials live on the host

v1 spec.yaml files continue to load unchanged via `LegacyXxx` shims on `specFile` and a polymorphic `credentialsField` decoder that handles both v1 (mapping with `sources:`) and v2 (sequence) shapes for the same `credentials:` YAML tag. A normalize pass folds v1 surfaces into the canonical fields with one deprecation warning per legacy block touched. Phase 6's schema cutover (later) removes the shims.

## Commits

- `d296aa1` feat(spec): Phase 3 — credentials[] + caps.network types (format migration only)
- `3f9d46c` chore(spec,tck): drop redundant Network canonical fields
- `17c3ad4` feat(bindings): add user credential bindings file format + loader (commit 10)
- `0a7ffff` fix(tck): allow routing-only credentials (empty `ApiKey.Name` when `Inject` non-empty)
- `48adfd7` fix(spec): sort legacy `serviceDomains` during normalize for determinism. Go map iteration order is randomized; without this, byte-identical specs produced different `Credentials[].ApiKey.Inject` orderings across loads and the sandboxes-side round-trip test flagged it as a regression.
- `ffc2e34` feat(bindings): `ParseValue` for raw + json-path discovery file parsers (commit 11a). Implements the parser dispatch declared by `DiscoveryFile.Parser`: `""` / `"raw"` returns content with trailing whitespace trimmed; `"json:a.b.c"` walks a dotted JSON path. Misses (key not present, non-string leaf) yield `ErrParserMiss` so the resolver can treat them as "no value here"; malformed parser specs yield `ErrParserMalformed` for the resolver to surface as a logged warning.
- `7e8eade` feat(bindings): allow binding entries with empty discovery list. Scenario 1c from the Phase 3 design discussion — value already in the sandboxes secret store, no separate discovery declaration needed — required the validator to accept bindings with just `allowedDomains`. The resolver consults the secret store before any discovery entries anyway, so making `discovery` optional is the clean expression of "trust these domains; the value is wherever `sbx secret set` put it."

## Scope discipline (important context for review)

Phase 3 is a **format migration only**. Pre-Phase-3 sandboxes passed credential field values through as opaque strings; semantic enforcement (format checks, security policy) happened at the runtime matcher / proxy / governance layer. This PR preserves that posture.

Several validation rules the original plan called for were removed during execution because they were **policy additions wearing a migration's clothes**:

- `apiKey.name` reserved-prefix rule (DASH_/SBX_/DOCKER_)
- `apiKey.inject[].format` must contain exactly one `%s`
- `oauth.passthrough = true` security-downgrade warning
- `provider` field deprecation-style warning
- `Credential.Service` lowercase-kebab pattern
- `apiKey.name` shell-identifier pattern
- `caps.network.allow/deny` entry format validation (P2 vs P3)
- Cross-block rule: `credentials[].apiKey.inject[].domain` must appear in `caps.network.allow` (RFC §818)

Each is tracked in `docs/specs/2026-05-29-unified-kit-spec-v2-deferred-ideas.md` on the sandboxes side, with the rule, why it has merit, why it was deferred, and how to add it later. Reopening any of them is fine — just under its own RFC, not under this migration.

`OAuthCredentialFile.Structure` (declarative JSON shape, security-motivated replacement for the existing free-form `Template` string) and `Credential.Provider` (forward-compat stub for the future provider registry) **do** ship in this PR — the first because removing an attack vector is the new format we want, the second because we want the full v2 spec shape in place.

## Test plan

- [x] `go test ./spec ./tck ./bindings -count=1` passes
- [x] `go vet ./...` clean
- [x] `gofmt -l ...` empty
- [x] v1 round-trip tests for credentials and caps.network shapes pass with deprecation warnings collected
- [x] New tests: `ParseValue` covers raw + json paths, misses, malformed specs (5 cases). Empty-discovery binding round-trips through `Load` + `Validate`.
- [x] Sandboxes-side wire baseline (`TestPhase3Baseline_KitCredentialResolution` in [docker/sandboxes#3267](https://github.com/docker/sandboxes/pull/3267)) stays byte-identical across all 7 sampled built-in kits — validates the wire-format-stability posture end-to-end.

### DIFF
--- a/bindings/bindings_test.go
+++ b/bindings/bindings_test.go
@@ -1,0 +1,134 @@
+package bindings
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad_ParsesExampleFile(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "credentials.yaml")
+	require.NoError(t, os.WriteFile(path, []byte(`
+bindings:
+  anthropic:
+    discovery:
+      - env: [ANTHROPIC_API_KEY]
+    allowedDomains: [api.anthropic.com]
+
+  github:
+    discovery:
+      - env: [GH_TOKEN, GITHUB_TOKEN]
+      - file:
+          path: ~/.config/gh/token
+    allowedDomains: [api.github.com, github.com]
+`), 0o600))
+
+	b, err := Load(path)
+	require.NoError(t, err)
+	require.NotNil(t, b)
+	require.Contains(t, b.Bindings, "anthropic")
+	require.Len(t, b.Bindings["anthropic"].Discovery, 1)
+	require.Equal(t, []string{"ANTHROPIC_API_KEY"}, b.Bindings["anthropic"].Discovery[0].Env)
+	require.Equal(t, []string{"api.anthropic.com"}, b.Bindings["anthropic"].AllowedDomains)
+
+	require.Contains(t, b.Bindings, "github")
+	require.Len(t, b.Bindings["github"].Discovery, 2)
+	require.Equal(t, []string{"GH_TOKEN", "GITHUB_TOKEN"}, b.Bindings["github"].Discovery[0].Env)
+	require.NotNil(t, b.Bindings["github"].Discovery[1].File)
+	require.Equal(t, "~/.config/gh/token", b.Bindings["github"].Discovery[1].File.Path)
+}
+
+func TestLoad_MissingFileIsError(t *testing.T) {
+	_, err := Load("/nonexistent/credentials.yaml")
+	require.Error(t, err)
+}
+
+func TestLoad_EmptyPathIsError(t *testing.T) {
+	_, err := Load("")
+	require.Error(t, err)
+}
+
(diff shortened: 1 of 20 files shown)


mdelapenya (dismissed):
LGTM! Only NIT is extending the existing test kits in the `testdata` dir, so they serves as reference and for testing purposes. But LGTM

bendebury:
@mdelapenya added a test kit for the V2 kit shape. Please take another look!

christopher-pluntke (approved):
LGTM based on Manu's review

tomyue-at-docker (approved):

PR #74, opened 2026-06-09
feat(spec): oauth.resourceHosts + v1 fold for OAuth-only services

bendebury:
## What

Give the v2 kit spec a first-class home for **OAuth resource hosts** so OAuth-only services can declare their routing without masquerading as a headerless `apiKey`.

- Add `oauth.resourceHosts []string` — the API hosts where the OAuth access-token bearer is used (e.g. `aiplatform.googleapis.com`), distinct from `tokenEndpoint.host` (where the token is refreshed).
- Add `Credential.RoutingHosts()` — the full routing set for a credential (apiKey inject domains ∪ oauth resource hosts ∪ oauth token endpoint), deduped + sorted.
- Normalizer: when folding the legacy standalone `oauth:` block, move a **routing-only** (nameless, headerless) `apiKey` — the shape v1 `network.serviceDomains` produces for an OAuth-only service — into `oauth.resourceHosts` (excluding the token endpoint), and drop the fake `apiKey`. Real api-key credentials are untouched.
- Fix the `Credential.OAuth` doc comment: precedence is **apiKey-wins** when both resolve (matches the runtime `OAuthEngine.GetAuthHeader`), not "OAuth wins".

## Why

Routing in v2 lived only under `apiKey.inject[].domain`, so the normalizer manufactured a degenerate headerless `apiKey` for OAuth-only services (e.g. `claude-vertex`). That confused downstream consumers (proxy routing, the credential binding gate) into treating OAuth routing as an API key. This change makes an OAuth credential declare both its domain-sets explicitly (resource hosts + token endpoint).

This is the spec-side prerequisite for the consumer changes in docker/sandboxes#3267 and for OAuth credential approval tracked in docker/sandboxes#3310.

## Tests

Unit tests for the new field/accessor and the normalizer fold (OAuth-only → `resourceHosts`; hybrid apiKey+OAuth keeps its `apiKey`). `go test ./spec/...` green.

Refs: docker/sandboxes#3267, docker/sandboxes#3310


mdelapenya (commented):
A few notes from reading through #74 — none are blockers.

### 1. Sort/dedupe in the OAuth-already-present merge branch (`spec/normalize.go:501-510`)

`sort.Strings(v2.ResourceHosts)` runs on the fold contribution *before* the
append into `c.OAuth.ResourceHosts`, so the merged list can end up unsorted
(and possibly duplicated) when `c.OAuth != nil`.

Trigger is narrow: a mixed spec with an existing
`Credentials[].OAuth.ResourceHosts`, a v1 standalone `oauth:` block, and a
routing-only apiKey for the same service. The function comment
("Deterministic order, independent of the upstream inject order") and the
assertion in `TestNormalize_OAuthOnlyService_DomainsMoveToResourceHosts`
both promise sorted output, so this looks like an oversight rather than
intentional.

Sketch:

```go
// After deciding set vs. merge:
target := s.Credentials.List[i].OAuth
target.ResourceHosts = append(target.ResourceHosts, /* moved domains */)
sort.Strings(target.ResourceHosts)
target.ResourceHosts = dedupSorted(target.ResourceHosts)
```

### 2. Test coverage for the merge path (`spec/normalize_test.go`)

Both new tests cover the `c.OAuth == nil` path. The `else`-merge branch added
in `fc17ed6` has no test. A case with existing
`Credentials[].OAuth{ResourceHosts: ["existing.com"]}` + a v1 standalone
`oauth:` block + a routing-only apiKey with `["new.com"]`, expecting a sorted
deduped result, would also catch (1).

### 3. (Nit) `spec/testdata/sample-agent-v2/spec.yaml`

The `workos` OAuth-only credential here doesn't use `resourceHosts`, and this
kit is documented as exercising the full v2 surface. A one-line
`resourceHosts: [api.workos.com]` would make the new field show up in the
reference example. Follow-up is fine.

---

For transparency, things I looked at and ruled out:

- **"apiKey takes precedence" comment fix** — merge commit notes
  `OAuthEngine.GetAuthHeader` was cross-checked.
- **`ProxyManaged` rename + `.gitignore` add** — both justified (sbx-cloud
  build break, stray local file).
- **`isRoutingOnlyApiKey` with empty `Inject`** — guarded upstream at
  `normalize.go:438`.
- **TCK coverage of `ResourceHosts`** — out of scope for a spec-only PR.

mdelapenya (approved):

PR #80, opened 2026-06-12
docs(kit-author): Phase 5 — sweep v1 spellings to v2 in user-facing docs

bendebury:
## Summary

Phase 5 (cleanup) doc sweep for the contrib repo. Translates stale v1 field-name spellings that the **kit-author–facing** docs still present as the current/recommended form into their v2 equivalents, so authors copy-paste v2.

Stacked on #79 (Phase 4 contrib). Docs only — no code, no spec/types change.

## What changed

- **Front-door** (`README.md`, `CONTRIBUTING.md`): the `spec.yaml` skeletons now show `schemaVersion: "2"` and the `caps.network` block; prose `network.allowedDomains` → `caps.network.allow`; the amp example `kind: agent` → `kind: sandbox`.
- **Per-kit READMEs** (amp, crush, hermes-agent, nanobot, nanoclaw, openclaw, pi, trivy): `kind: agent` → `kind: sandbox`; `network`/`serviceDomains`/`serviceAuth`/`credentials.sources`/`proxyManaged` prose → the v2 `credentials[]` + `caps.network` shapes (discovery half moves to the user-side bindings file).
- **`spec-anatomy.md`**: tagged the three **parsed-but-unwired** v2 fields (`entrypoint.ttyArgs`, `entrypoint.pipeMode`, `apiKey.inject[].username`) as **P3 — pending sbx support**, using the doc's existing convention, so they're documented as part of the v2 *shape* without implying they're wired yet.
- **Install-lane reconciliation** (`spec-anatomy.md`, `v1-migration.md`): the `settings` → `commands.initFiles` guidance now also points credential/mode-dependent files (branching on `SBX_CRED_<SERVICE>_MODE`) at `commands.install`, matching how Phase 4 actually landed.

## What was deliberately NOT touched

Intentional v1 references that document the migration itself were left as-is: `v1-migration.md`, the `lifecycle.md` normalization mapping, `pitfalls.md`'s legacy-fields list, `scripts/README.md`'s migrate-tool table, and the `proxyManaged`/`settings` **removal** notes.

```release-note
NONE
```

PR #90, opened 2026-06-12
feat(spec): apiKey.proxyManaged — per-credential in-container sentinel marker

bendebury:
## Summary

Re-expresses the removed v1 `environment.proxyManaged` list as a per-credential **`apiKey.proxyManaged: bool`** (default `false`). `normalize` folds v1 `proxyManaged` membership onto the marker; the migrate script emits it. This is the prerequisite for migrating the built-in agent specs (docker/sandboxes) to schemaVersion 2 — the engine will derive the in-container sentinel set from this marker instead of the v1 shim.

Stacked on the Phase 5 doc-sweep branch (#80).

## Why

A full-Artifact gap sweep of the 12 built-in specs showed `Environment.ProxyManaged` is the *only* field that differs v1→v2, and it's a **curated, non-derivable** authoring decision — "which credential env vars are exposed in-container as sentinels." It is **not** redundant with `apiKey.name`: `shell` proxy-manages the provider keys but not `github`; the agent kits (`claude`/`codex`/…) proxy-manage *nothing* despite having an `apiKey`. So v2 must represent it explicitly.

## For review (please flag if you object)

1. **Field name `proxyManaged` + the `false` default** — in-container exposure is opt-in (conservative/secure), vs. the RFC's implicit "`apiKey.name` is always in-container." The agent kits rely on it staying off.
2. **`aliases` / multi-name in-container exposure is intentionally NOT added (YAGNI).** The only case across the built-ins is `shell` exposing google under both `GOOGLE_API_KEY` and `GEMINI_API_KEY` (blanket coverage of a generic shell). We expose the canonical name only; a `GEMINI_API_KEY`-only tool (the `gemini` CLI) has its own dedicated kit. Purely additive later if a real need appears.

```release-note
NONE
```

### DIFF
--- a/scripts/migrate_v1_to_v2_test.go
+++ b/scripts/migrate_v1_to_v2_test.go
@@ -224,3 +224,28 @@ func TestMigrate_Idempotent(t *testing.T) {
 		t.Errorf("re-migrating canonical v2 output should be a no-op; got changes %v", changes)
 	}
 }
+
+// TestMigrateEmitsProxyManaged confirms a v1 environment.proxyManaged entry
+// folds onto credentials[].apiKey.proxyManaged: true in the migrated output.
+func TestMigrateEmitsProxyManaged(t *testing.T) {
+	v1 := []byte(`schemaVersion: "1"
+kind: agent
+name: t
+agent: {image: x}
+network:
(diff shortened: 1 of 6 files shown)


mdelapenya (approved):
LGTM

PR #92, opened 2026-06-13
fix(junie): add github's release-assets allowed domain and CODEOWNERS

mdelapenya:
- **fix(junie): include release-assets allowed domain**
- **chore(junie): add CODEOWNERS**


mdelapenya:
@preigile I added all the domains that the kit needs to work in an air-gaped environment (default policy: `deny-all`). Could you please take a look?

ilopezluna (approved):

PR #96, opened 2026-06-15
feat(bindings): forward-compat for named variants + workspace associations

bendebury:
Additive forward-compat for the user credential bindings file (`~/.config/sbx/credentials.yaml`), in support of the v2 binding-authority work (spec D11). **No resolution/discovery behavior change.**

## What
- Add `UserBindings.Remembered map[string]map[string]string` (RFC P2 workspace associations) with `yaml:"remembered,omitempty"`, so the section round-trips through load → `yaml.Marshal` → load instead of being silently dropped when the (forthcoming) sandboxes-side consent flow rewrites the file.
- Doc-comment update noting `service@variant` binding keys are preserved verbatim.
- Tests: round-trip preservation of both a `remembered:` section and a `service@variant` key; characterization test locking `@variant`-key and domains-only binding validation tolerance.

## Why now
Named variants / workspace associations are **not implemented** here — this only guarantees lossless preservation so the sandboxes binding-authority branch (which rewrites this file) can't destroy a user's hand-written config. That
branch will pin `go.mod` to this commit.

## Out of scope
Variant selection/parsing, `remembered` semantics validation, any discovery change.


mdelapenya (dismissed):
LGTM

@bendebury should we move this bindings package inside the spec one? I think so

mdelapenya (approved):

PR #98, opened 2026-06-16
Add terraform ops kit

falconcr:
### Add Terraform Ops Kit

A mixin kit that adds a complete Terraform infrastructure-as-code toolkit to any sbx agent (Claude, Gemini, Copilot, Shell).

### What it does
Pre-installs terraform, terragrunt, tflint, checkov, infracost, and aws-cli in the sandbox at creation time. Enables AI agents to autonomously plan, validate, lint, security-scan, and cost-estimate Terraform infrastructure.

Spec choices worth flagging

- Terragrunt pinned to v0.59.3 — GitHub API is rate-limited inside sandboxes (no auth token), making latest detection unreliable. Pinning avoids that.
- Infracost pinned to v0.10.31 — The upstream install.sh script redirects to a new CLI repository for v1+. Direct binary download from the stable v0.x tag is more reliable.
- Checkov via apt (python3-checkov) — pip3 triggers PEP 668 externally-managed-environment errors on Ubuntu 24.04. 

### APT avoids the conflict. Falls back gracefully if the package is not in distro.
- Ubuntu release detection — Uses /etc/os-release instead of lsb_release (not available in shell-docker image).
- curl over wget — shell-docker image ships curl but not wget.

### Test plan
- sbx kit validate — VALID
- Created sandbox with sbx create --kit ./terraform-ops-kit/ shell /tmp/tf-test — all 6 install commands executed successfully
- Verified terraform, terragrunt, tflint, infracost, and aws-cli all report their versions post-install

### Files
terraform-ops-kit/
├── spec.yaml  
└── README.md    
---

## Maintainer updates (schemaVersion 2)

- Migrated to spec `schemaVersion: "2"` (every other kit in the repo is on v2 as of #192).
- Fixed checkov: apt's `python3-checkov` doesn't exist in the base image's distro, so the actual install always fell back silently — switched to `pip install --break-system-packages --ignore-installed checkov`, matching upstream's own documented install method.
- Fixed tflint: replaced the curl-piped install script with a version-pinned, SHA256-verified binary download (matches the `task`/`trivy` precedent already in this repo).
- Fixed the AWS Credentials section: `sbx create --mount` is not a real flag. Replaced with the actual `--env`/`-e` and `--env-file` flags for passing credentials into the sandbox.
- Hardened all curl downloads (terragrunt, tflint, infracost, aws-cli, HashiCorp gpg key) with `--retry 3 --retry-delay 2 --retry-all-errors` — found via a real transient DNS blip on `github.com` during TCK, reproduced and confirmed transient before applying the fix.
- Added the OCI-artifact usage example and stripped `$ ` prompt prefixes from README code blocks (repo-wide convention from #214).


kiview (approved):

mdelapenya:
@falconcr question on scope: `terraform`, `terragrunt`, `tflint`, `checkov`, and `infracost` are all cloud-provider-agnostic, but `aws-cli` only matters if you're on AWS. Would it make sense to split this into two kits — a provider-agnostic "terraform ops" kit with the five IaC tools, and `aws-cli` as a separate companion mixin? That way GCP/Azure Terraform users aren't paying the install cost and allow-listing `awscli.amazonaws.com` for a tool they'll never use. Not a blocker, just want to hear your take before this lands — happy to help with the split if you think it's worth it.

falconcr:
> question on scope: `terraform`, `terragrunt`, `tflint`, `checkov`, and `infracost` are all cloud-provider-agnostic, but `aws-cli` only matters if you're on AWS. Would it make sense to split this into two kits — a provider-agnostic "terraform ops" kit with the five IaC tools, and `aws-cli` as a separate companion mixin? That way GCP/Azure Terraform users aren't paying the install cost and allow-listing `awscli.amazonaws.com` for a tool they'll never use. Not a blocker, just want to hear your take before this lands — happy to help with the split if you think it's worth it.

Make sense, feel free to add or update it.

PR #101, opened 2026-06-18
feat(spec): add streaming path so loaders don't buffer files/ into memory

### LINKED ISSUE #97: `spec`: add a streaming path so loaders don't buffer all of `files/` into memory
## Summary

The `spec` loaders eagerly read every file under a kit's `files/` tree into
memory (`ArtifactFile.Content []byte`) before the caller touches anything.
That's fine for the laptop CLI path but wasteful for server-side consumers
that only need metadata or only want to forward bytes onward. Propose adding
an **additive, opt-in streaming path**: kit metadata is loaded eagerly, and
each file's content is read through a single canonical `Open()` method that
streams on demand. Content is **never buffered or cached** by the loader on
this path. The existing eager API and `ArtifactFile.Content` are left
untouched.

## Motivation

1. **Server-side orchestration / fleet deployment.** An agentic-workflow
   orchestrator that deploys kits to a fleet of remote sandboxes currently
   has to load each kit's entire `files/` tree into the orchestrator process
   just to push the bytes onward. With streaming, the orchestrator copies
   each file straight from source to destination and memory stays flat
   regardless of `kits × files × size`.
2. **Large static assets.** Kits that bundle larger files (config trees,
   model-weight stubs, fixtures) force the whole payload into RAM even when
   the caller only needs to read `spec.yaml` and enumerate file metadata.
3. **Progressive transfer.** A future `sbx` server or kit registry can copy
   `io.Reader` → destination as bytes arrive, instead of buffering each file
   fully first.

## Current behavior (verified)

In `spec/artifact.go`:

- `LoadFromDirectory(dir)` and `LoadFromFS(fsys, dir)` parse `spec.yaml`, then
  call `collectFilesFromDir` / `collectFilesFromFS`, which walk `files/home`
  and `files/workspace` and `os.ReadFile` / `fs.ReadFile` **every** file fully
  into `ArtifactFile.Content []byte`.
- `LoadFromBytes(yamlBytes)` parses only `spec.yaml` and leaves `Files` unset
  (the building block the out-of-repo v2 OCI puller uses).

The in-memory shape (`spec/types.go`):

```go
type ArtifactFile struct {
    RelativePath string // path relative to target root (home/ or workspace/)
    Target       string // "home" or "workspace"
    Mode         int64
    Content      []byte // raw file bytes — fully buffered today
    Size         int64
}
```

`Size` is already populated independently of `Content`, so file metadata is
available without the bytes.

## Proposal

Make `Open()` the canonical way to read a file's content — regardless of how
the artifact was loaded — and add a streaming loader that populates file
metadata eagerly while deferring the bytes.

```go
type ArtifactFile struct {
    RelativePath string `json:"relativePath"`
    Target       string `json:"target"`
    Mode         int64  `json:"mode"`
    Content      []byte `json:"content"` // eager bytes; nil on the streaming path
    Size         int64  `json:"size"`

    // ContentSource streams this file's content on demand. Streaming loaders
    // set it; eager loaders leave it nil. Excluded from JSON — a stream is
    // process-local and not serializable (see Open and Materialize).
    //
    // Contract: ContentSource may be called multiple times and in any order;
    // each call returns an independent reader positioned at the start of the
    // file (random access, repeatable).
    ContentSource func() (io.ReadCloser, error) `json:"-"`
}

// Open is the canonical way to read a file's content. It serves Content when
// present, otherwise calls ContentSource. Callers must Close the reader.
func (f *ArtifactFile) Open() (io.ReadCloser, error)

// OpenFromDirectory / OpenFromFS load metadata eagerly (RelativePath/Target/
// Mode/Size) and set ContentSource to stream each file on demand; Content is
// left nil. The source must remain available for as long as the Artifact is
// in use.
func OpenFromDirectory(dir string) (*Artifact, error)
func OpenFromFS(fsys fs.FS, dir string) (*Artifact, error)
```

Consumers read uniformly, whether the artifact was loaded eagerly or streamed:

```go
rc, err := f.Open()
if err != nil { /* ... */ }
defer rc.Close()
io.Copy(dst, rc) // never holds the whole file in memory on the streaming path
```

Discriminator invariant: eager loaders set `Content` non-nil (possibly empty)
and leave `ContentSource` nil; streaming loaders leave `Content` nil and set
`ContentSource`.

**Random access is the loader's responsibility.** `ContentSource` must support
repeated, any-order reads. For a local directory this is free. For sources that
are naturally single-pass front-to-back streams — notably an OCI tar layer —
implementers should consider this carefully: backing the source with a local
on-disk (pull-through) cache is a sensible way to satisfy the contract, so
reading one file, re-reading, or reading files out of order doesn't re-fetch
the whole stream. The OCI puller lives outside this repo (today built on
`LoadFromBytes`); this issue only fixes the contract it must meet — the puller
should populate `ContentSource` rather than `io.ReadAll`-ing every entry into
`Content`.

**Serialization bridge.** Add `Artifact.Materialize()` to read every streamed
file (`Content == nil`, `ContentSource != nil`) fully into `Content` via
`io.ReadAll`, making the artifact JSON-serializable again. Call it before
marshalling at a serialization boundary.

```go
func (a *Artifact) Materialize() error
```

**Safety checks.** Loaders run path-safety checks eagerly at enumeration time —
local symlink-escape, OCI tar path/type/mode — independent of when bytes are
read, so streamed content is gated by the same validation.

## Backward compatibility (hard requirement)

- `LoadFromDirectory`, `LoadFromFS`, and `LoadFromBytes` keep their current
  signatures and behavior — eager loads still fully populate
  `ArtifactFile.Content`, and `ContentSource` is `nil`.
- `ArtifactFile.Content []byte` stays as-is. Existing consumers that read
  `.Content` keep working unchanged. The new `ContentSource` field is tagged
  `json:"-"`, so existing `json.Marshal` / `Unmarshal` of the kit artifact
  (e.g. the `sandboxd` add-kit API) is unaffected — an exported func field
  would otherwise break `Marshal`.
- `Open()`, `OpenFromDirectory` / `OpenFromFS`, and `Materialize()` are purely
  **additive**. No existing caller is forced to migrate; `Open()` is the
  recommended read API going forward.

## Affected code surface

- **`spec` package** — new `OpenFromDirectory` / `OpenFromFS` alongside
  `collectFilesFromDir` / `collectFilesFromFS`; the `ContentSource` field, the
  `Open()` method, and `Materialize()`. The symlink-escape check that
  `collectFilesFromDir` performs during the eager walk must also run at
  enumeration time on the streaming path (and the source should open the
  validated real path).
- **`spec/validate.go`** — `ValidateArtifact` iterates `Files` but only checks
  `Target` / `RelativePath` (path-safety); it does **not** read `Content`, so
  validation already works for a metadata-only / streaming load.
- **`tck` package** — the only real consumer of `.Content` is `tck/suite.go`
  (`container.CopyToContainer(ctx, f.Content, ...)`); `tck/derive.go` reads
  `Files` but only uses `Target` / `RelativePath`. The TCK can move to
  `f.Open()` as the canonical read.
- **Serialization boundaries** (e.g. the out-of-repo `sandboxd` add-kit API
  that JSON-marshals the kit artifact) call `Materialize()` before marshalling
  a streamed artifact.

## Non-goals

- Changing the on-disk kit format or `spec.yaml` schema.
- Removing or deprecating the eager `Content []byte` API.
- Making serialization boundaries stream — the `sandboxd` add-kit API marshals
  `Files[].Content` inline as JSON; a streamed artifact must be materialized
  first (see `Materialize`).

simonferquel-clanker:
Closes #97

## What changed

- **`ArtifactFile.ContentSource`** — new `func() (io.ReadCloser, error)` field tagged `json:"-"` so `json.Marshal` never touches the un-serializable function value.
- **`ArtifactFile.Open()`** — canonical read method: returns a reader backed by `Content` when non-nil (eager path), calls `ContentSource` (streaming path), or errors if neither is set.
- **`OpenFromDirectory` / `OpenFromFS`** — streaming loaders that enumerate metadata and run all safety checks (symlink-escape, path validation) eagerly, but defer content reads to `Open()`.
- Shared enumeration helpers `enumerateDirFiles` / `enumerateFSFiles` are used by both the eager and streaming loaders so the symlink-escape check is never duplicated.
- **`Artifact.Materialize()`** — converts a streaming artifact to the eager form by reading all `ContentSource` values into `Content`. Idempotent, checks the `Close` error, and clears `ContentSource` to `nil` afterwards.
- **tck**: `NewSuiteFromDir` now uses `OpenFromDirectory` (exercises the streaming path); `RunContainerTests` reads bytes via `f.Open()` + `io.ReadAll` instead of `f.Content` directly.

All existing exported signatures and eager-path semantics are unchanged (byte-for-byte backward compatible).

## Tests

- `spec/streaming_test.go` (new): full matrix — `Open()` fallback logic, metadata parity, random/repeated access, symlink-escape safety, empty-file handling, `OpenFromFS` parity with MapFS and `os.DirFS`, `Materialize` scenarios, JSON safety (`json:"-"` proves correct), and `ValidateArtifact` on streaming artifacts.
- All existing tests pass; `go test -short ./...` is clean.

### DIFF
--- a/spec/artifact.go
+++ b/spec/artifact.go
@@ -3,6 +3,7 @@ package spec
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -19,7 +20,9 @@ const (
 	filesDirWork    = "files/workspace"
 )
 
-// LoadFromDirectory loads a kit artifact from a local directory.
+// LoadFromDirectory loads a kit artifact from a local directory. All file
+// content is read into memory (eager path). The returned artifact is fully
+// validated.
 func LoadFromDirectory(dir string) (*Artifact, error) {
 	info, err := os.Stat(dir)
 	if err != nil {
@@ -38,10 +41,26 @@ func LoadFromDirectory(dir string) (*Artifact, error) {
 		return nil, err
 	}
 
-	artifact.Files, err = collectFilesFromDir(dir)
+	entries, err := enumerateDirFiles(dir)
 	if err != nil {
 		return nil, fmt.Errorf("artifact files: %w", err)
 	}
+
+	files := make([]ArtifactFile, 0, len(entries))
+	for _, e := range entries {
+		data, err := os.ReadFile(e.realPath)
+		if err != nil {
+			return nil, fmt.Errorf("artifact files: read file %s: %w", e.realPath, err)
+		}
+		files = append(files, ArtifactFile{
+			RelativePath: e.relativePath,
+			Target:       e.target,
+			Mode:         e.mode,
+			Content:      data,
+			Size:         e.size,
+		})
+	}
+	artifact.Files = files
 
 	if err := ValidateArtifact(artifact); err != nil {
 		return nil, err
@@ -50,7 +69,62 @@ func LoadFromDirectory(dir string) (*Artifact, error) {
 	return artifact, nil
 }
 
-// LoadFromFS loads a kit artifact from an fs.FS (e.g., embed.FS or os.DirFS).
+// OpenFromDirectory opens a kit artifact from a local directory using the
+// streaming path. Metadata is enumerated and all safety checks (symlink
+// escape, path validation) are performed eagerly at load time, but file
+// content is not read into memory. Each ArtifactFile.Open call opens the
+// file on demand, allowing callers to process one file at a time.
+//
+// The returned artifact is fully validated. Streaming files have Content nil
+// and ContentSource non-nil; use ArtifactFile.Open to read content, or call
+// Artifact.Materialize to populate Content for all files at once.
+func OpenFromDirectory(dir string) (*Artifact, error) {
+	info, err := os.Stat(dir)
+	if err != nil {
+		return nil, fmt.Errorf("artifact directory: %w", err)
+	}
+	if !info.IsDir() {
+		return nil, fmt.Errorf("artifact: %q is not a directory", dir)
+	}
+
+	readFile := func(name string) ([]byte, error) {
+		return os.ReadFile(filepath.Join(dir, name))
+	}
+
+	artifact, err := parseArtifact(readFile)
+	if err != nil {
+		return nil, err
+	}
+
+	entries, err := enumerateDirFiles(dir)
+	if err != nil {
+		return nil, fmt.Errorf("artifact files: %w", err)
+	}
+
+	files := make([]ArtifactFile, 0, len(entries))
+	for _, e := range entries {
+		files = append(files, ArtifactFile{
+			RelativePath: e.relativePath,
+			Target:       e.target,
+			Mode:         e.mode,
+			Size:         e.size,
+			ContentSource: func() (io.ReadCloser, error) {
+				return os.Open(e.realPath)
+			},
+		})
+	}
+	artifact.Files = files
+
+	if err := ValidateArtifact(artifact); err != nil {
+		return nil, err
+	}
+
+	return artifact, nil
+}
+
+// LoadFromFS loads a kit artifact from an fs.FS (e.g., embed.FS or
+// os.DirFS). All file content is read into memory (eager path). The
+// returned artifact is fully validated.
 func LoadFromFS(fsys fs.FS, dir string) (*Artifact, error) {
 	dir = filepath.ToSlash(dir)
 
@@ -63,10 +137,72 @@ func LoadFromFS(fsys fs.FS, dir string) (*Artifact, error) {
 		return nil, err
 	}
 
-	artifact.Files, err = collectFilesFromFS(fsys, dir)
+	entries, err := enumerateFSFiles(fsys, dir)
 	if err != nil {
 		return nil, fmt.Errorf("artifact files: %w", err)
 	}
+
+	files := make([]ArtifactFile, 0, len(entries))
+	for _, e := range entries {
+		data, err := fs.ReadFile(fsys, e.fsPath)
+		if err != nil {
+			return nil, fmt.Errorf("artifact files: read file %s: %w", e.fsPath, err)
+		}
+		files = append(files, ArtifactFile{
+			RelativePath: e.relativePath,
+			Target:       e.target,
+			Mode:         e.mode,
+			Content:      data,
+			Size:         int64(len(data)),
+		})
+	}
+	artifact.Files = files
+
+	if err := ValidateArtifact(artifact); err != nil {
+		return nil, err
+	}
+
+	return artifact, nil
+}
+
+// OpenFromFS opens a kit artifact from an fs.FS (e.g., embed.FS or
+// os.DirFS) using the streaming path. Metadata is enumerated eagerly but
+// file content is not read into memory. Each ArtifactFile.Open call opens
+// the file on demand via the FS.
+//
+// The returned artifact is fully validated. Streaming files have Content nil
+// and ContentSource non-nil; use ArtifactFile.Open to read content, or call
+// Artifact.Materialize to populate Content for all files at once.
+func OpenFromFS(fsys fs.FS, dir string) (*Artifact, error) {
+	dir = filepath.ToSlash(dir)
+
+	readFile := func(name string) ([]byte, error) {
+		return fs.ReadFile(fsys, filepath.ToSlash(filepath.Join(dir, name)))
+	}
+
+	artifact, err := parseArtifact(readFile)
+	if err != nil {
+		return nil, err
+	}
+
+	entries, err := enumerateFSFiles(fsys, dir)
+	if err != nil {
+		return nil, fmt.Errorf("artifact files: %w", err)
+	}
+
+	files := make([]ArtifactFile, 0, len(entries))
+	for _, e := range entries {
+		files = append(files, ArtifactFile{
+			RelativePath: e.relativePath,
+			Target:       e.target,
+			Mode:         e.mode,
+			Size:         e.size,
+			ContentSource: func() (io.ReadCloser, error) {
+				return fsys.Open(e.fsPath)
+			},
+		})
+	}
+	artifact.Files = files
 
 	if err := ValidateArtifact(artifact); err != nil {
 		return nil, err
@@ -87,6 +223,39 @@ func LoadFromFS(fsys fs.FS, dir string) (*Artifact, error) {
 // has no notion of a file source.
 func LoadFromBytes(yamlBytes []byte) (*Artifact, error) {
 	return parseArtifactBytes(yamlBytes)
+}
+
+// Materialize reads any streamed files (Content == nil, ContentSource set)
+// into Content and updates Size. Files already on the eager path (Content
+// non-nil) are left unchanged, making the call idempotent. ContentSource is
+// cleared to nil after the content is read so the file is unambiguously on
+// the eager path afterwards. Returns an error if any file has neither Content
+// nor ContentSource.
+func (a *Artifact) Materialize() error {
+	for i := range a.Files {
+		f := &a.Files[i]
+		if f.Content != nil {
+			continue // already eager; nothing to do
+		}
+		if f.ContentSource == nil {
+			return fmt.Errorf("artifact: file %q has neither Content nor ContentSource", f.RelativePath)
+		}
+		rc, err := f.ContentSource()
+		if err != nil {
+			return fmt.Errorf("artifact: open %q: %w", f.RelativePath, err)
+		}
+		data, readErr := io.ReadAll(rc)
+		if cerr := rc.Close(); readErr == nil && cerr != nil {
+			return fmt.Errorf("artifact: close %q: %w", f.RelativePath, cerr)
+		}
+		if readErr != nil {
+			return fmt.Errorf("artifact: read %q: %w", f.RelativePath, readErr)
+		}
+		f.Content = data
+		f.Size = int64(len(data))
+		f.ContentSource = nil // file is now fully on the eager path
+	}
+	return nil
 }
 
 // parseArtifact reads spec.yaml (or spec.yml as fallback) via readFile and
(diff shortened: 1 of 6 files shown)


tck/suite.go
```diff
@@ -86,7 +86,12 @@ func (s *Suite) RunContainerTests(t *testing.T) {
 			require.NoError(t, err)
 			require.Equal(t, 0, code, "mkdir -p %s failed", parentDir)
 
-			err = container.CopyToContainer(ctx, f.Content, containerPath, f.Mode)
+			rc, err := f.Open()
+			require.NoError(t, err, "failed to open %s for container copy", containerPath)
+			fileBytes, err := io.ReadAll(rc)
+			_ = rc.Close()
+			require.NoError(t, err, "failed to read %s for container copy", containerPath)
+			err = container.CopyToContainer(ctx, fileBytes, containerPath, f.Mode)
```
simonferquel, line 95: Is'nt there a way to stream the content to the container rather than reading the file fully ?

simonferquel-clanker:
The testcontainers `Container` interface only exposes `CopyToContainer([]byte)`, so buffering one file at a time at the copy boundary is unavoidable here. Real consumers of the streaming API (e.g. the OCI puller in sandboxlib) will call `Open()` and pipe the reader directly without an intermediate buffer — that's where the win is. The TCK is kept simple.

spec/artifact.go
```diff
@@ -89,6 +225,39 @@ func LoadFromBytes(yamlBytes []byte) (*Artifact, error) {
 	return parseArtifactBytes(yamlBytes)
 }
 
+// Materialize reads any streamed files (Content == nil, ContentSource set)
+// into Content and updates Size. Files already on the eager path (Content
+// non-nil) are left unchanged, making the call idempotent. ContentSource is
+// cleared to nil after the content is read so the file is unambiguously on
+// the eager path afterwards. Returns an error if any file has neither Content
+// nor ContentSource.
+func (a *Artifact) Materialize() error {
```
mdelapenya, line 238: Nice job splitting enumeration from content loading here 🙌

One edge case on `Materialize`: it mutates files in place and `return`s on the first error, so a mid-loop failure leaves the artifact **half-eager / half-streaming with no rollback**. Combined with the JSON tags (`Content` is `json:"content"` with no `omitempty`, `ContentSource` is `json:"-"`), a still-streaming file then marshals as `"content": null` — so a caller that serializes the artifact after a failed `Materialize` silently loses those files' bytes, with no error anywhere.

Suggested fix — make it all-or-nothing by staging reads and committing only on full success:

```go
func (a *Artifact) Materialize() error {
	staged := make([][]byte, len(a.Files))
	for i := range a.Files {
		f := &a.Files[i]
		if f.Content != nil {
			continue
		}
		if f.ContentSource == nil {
			return fmt.Errorf("artifact: file %q has neither Content nor ContentSource", f.RelativePath)
		}
		rc, err := f.ContentSource()
		if err != nil {
			return fmt.Errorf("artifact: open %q: %w", f.RelativePath, err)
		}
		data, readErr := io.ReadAll(rc)
		if cerr := rc.Close(); readErr == nil && cerr != nil {
			return fmt.Errorf("artifact: close %q: %w", f.RelativePath, cerr)
		}
		if readErr != nil {
			return fmt.Errorf("artifact: read %q: %w", f.RelativePath, readErr)
		}
		staged[i] = data
	}
	for i := range a.Files {
		if staged[i] == nil {
			continue // already eager / untouched
		}
		a.Files[i].Content = staged[i]
		a.Files[i].Size = int64(len(staged[i]))
		a.Files[i].ContentSource = nil
	}
	return nil
}
```

Test that fails today and passes with the fix:

```go
func TestMaterialize_PartialFailureIsAtomic(t *testing.T) {
	a := &Artifact{Files: []ArtifactFile{
		{RelativePath: "a.txt", Target: TargetHome, ContentSource: func() (io.ReadCloser, error) {
			return io.NopCloser(bytes.NewReader([]byte("first"))), nil
		}},
		{RelativePath: "b.txt", Target: TargetHome, ContentSource: func() (io.ReadCloser, error) {
			return nil, errors.New("backing file vanished after enumeration")
		}},
	}}

	require.Error(t, a.Materialize())

	// a.txt must NOT be left eager while b.txt failed: otherwise a later
	// json.Marshal emits a.txt's bytes but b.txt as "content": null, silently
	// dropping b.txt's payload with no error.
	require.Nil(t, a.Files[0].Content, "no file may be left eager after a failed Materialize")
}
```

spec/artifact.go
```diff
@@ -63,11 +137,73 @@ func LoadFromFS(fsys fs.FS, dir string) (*Artifact, error) {
...
+		data, err := fs.ReadFile(fsys, e.fsPath)
+		if err != nil {
+			return nil, fmt.Errorf("artifact files: read file %s: %w", e.fsPath, err)
+		}
+		files = append(files, ArtifactFile{
+			RelativePath: e.relativePath,
+			Target:       e.target,
+			Mode:         e.mode,
+			Content:      data,
+			Size:         int64(len(data)),
```
mdelapenya: Small parity gap here: the eager FS path sets `Size` from the bytes it read (`int64(len(data))`), while the streaming path **and both directory paths** use the enumerated `FileInfo.Size()` (`e.size`). They agree for OS dirs / `embed.FS` / `MapFS`, but diverge for any `fs.FS` where `Info().Size()` ≠ the byte length — and since `enumerateFSFiles` already computed it, `e.size` is effectively dead on this path.

Suggested fix — use the enumerated size like the other three loaders:

```go
		files = append(files, ArtifactFile{
			RelativePath: e.relativePath,
			Target:       e.target,
			Mode:         e.mode,
			Content:      data,
			Size:         e.size, // was int64(len(data)); match dir + streaming paths
		})
```

`TestOpenFromFS_Parity` doesn't assert `Size` today, so this slips through. Worth adding to the parity loop:

```go
require.Equal(t, ef.Size, sf.Size, "Size mismatch for %s", key)
```

…plus a focused test with an `fs.FS` whose reported size lies (fails today, passes after the fix):

```go
// lyingSizeFS reports a FileInfo.Size() that differs from the bytes Open() yields.
type lyingSizeFS struct {
	fstest.MapFS
	lie map[string]int64
}

func (l lyingSizeFS) ReadDir(name string) ([]fs.DirEntry, error) {
	ents, err := l.MapFS.ReadDir(name)
	if err != nil {
		return nil, err
	}
	for i, e := range ents {
		if sz, ok := l.lie[path.Join(name, e.Name())]; ok {
			ents[i] = lyingDirEntry{DirEntry: e, size: sz}
		}
	}
	return ents, nil
}

type lyingDirEntry struct {
	fs.DirEntry
	size int64
}

func (e lyingDirEntry) Info() (fs.FileInfo, error) {
	info, err := e.DirEntry.Info()
	if err != nil {
		return nil, err
	}
	return lyingFileInfo{FileInfo: info, size: e.size}, nil
}

type lyingFileInfo struct {
	fs.FileInfo
	size int64
}

func (i lyingFileInfo) Size() int64 { return i.size }

func TestOpenFromFS_SizeMatchesLoadFromFS(t *testing.T) {
	const data = "hello from fs" // 13 bytes
	fsys := lyingSizeFS{
		MapFS: fstest.MapFS{
			"mykit/spec.yaml":            {Data: []byte(minimalMixinYAML)},
			"mykit/files/home/hello.txt": {Data: []byte(data), Mode: 0o644},
		},
		lie: map[string]int64{"mykit/files/home/hello.txt": 9999},
	}

	eager, err := LoadFromFS(fsys, "mykit")
	require.NoError(t, err)
	streaming, err := OpenFromFS(fsys, "mykit")
	require.NoError(t, err)

	require.Equal(t, eager.Files[0].Size, streaming.Files[0].Size) // 13 vs 9999 today
}
```

mdelapenya (commented):
Really nice change — the enumerate-then-load split is clean, the `Open`/`Materialize` contract reads well, and the test coverage is thorough 👏

Two things worth tightening before merge, both around the eager↔streaming metadata/content contract. Neither bites on the happy path; they show up on error and edge-case `fs.FS` inputs. Details inline, each with a suggested fix + a test you can drop in. 🐳

simonferquel-clanker:
> Nice job splitting enumeration from content loading here 🙌 One edge case on `Materialize`: it mutates files in place and returns on the first error, so a mid-loop failure leaves the artifact half-eager / half-streaming with no rollback.

Fixed in 941de76. `Materialize` now runs two phases: reads are staged into a local slice first, then committed to `a.Files` only after every read succeeds. A mid-loop failure leaves the artifact entirely unchanged. `TestMaterialize_PartialFailureIsAtomic` covers the exact scenario.

simonferquel-clanker:
> Small parity gap: the eager FS path sets Size from the bytes it read (int64(len(data))), while the streaming path uses the enumerated FileInfo.Size() (e.size).

Fixed in 941de76. `LoadFromFS` now takes `Size` from `e.size` (matching `OpenFromFS`), so both loaders report the same `Size` for the same file. `TestOpenFromFS_SizeMatchesLoadFromFS` uses a lying `fs.FS` (declared size 9999, actual bytes 13) to pin this. Also added a `Size` assertion to the `TestOpenFromFS_Parity/mapfs` loop.

simonferquel-clanker:
Re Copilot comments on loop-variable capture in `OpenFromDirectory` (line 116) and `OpenFromFS` (line 205): this module requires Go 1.26.2 (`go.mod`). Since Go 1.22 each `for range` iteration gets its own variable, so the closures already capture independent copies of `e`. No shadowing needed.

simonferquel-clanker:
Re remaining Copilot findings — all fixed in earlier commits:
- **Size from stat vs len(data)** (line 61): `LoadFromDirectory` now uses `int64(len(data))` — fixed in 12e46aa, covered by `TestLoadFromDirectory_SizeEqualsContentLen`.
- **realPath not absolute** (line 396/409): `enumerateDirFiles` now stores `absReal` instead of `realPath` — fixed in 12e46aa, covered by `TestOpenFromDirectory_CWDIndependence`.
- **rc.Close() ignored in suite.go**: now captured and asserted — fixed in 12e46aa.

mdelapenya (approved):

PR #109, opened 2026-06-23
Update NanoClaw sandbox kit

amit-shafnir:
## Summary

Update the NanoClaw kit to use the prebuilt sandbox-host image flow.

This `spec.yaml` is based on the NanoClaw sandbox spec from #33 by @ealeyner:
https://github.com/docker/sbx-kits-contrib/pull/33

That PR established the right direction for NanoClaw in sbx: a `kind: sandbox`
kit backed by a prebuilt host image, instead of installing and building NanoClaw
during sandbox startup.

## Changes

This PR keeps that model, but updates it to the current NanoClaw packaging:

- Uses `docker.io/nanoco/nanoclaw:sbx-claude-alpha` as the sandbox host image.
- Starts NanoClaw through the image-provided `/usr/local/bin/nanoclaw-start`.
- Keeps image build logic out of this repo; this PR only updates the kit spec and README.
- Uses NanoClaw’s normal OneCLI credential flow.
- Declares the current outbound network policy with `caps.network.allow`.
- Publishes the webhook, OneCLI dashboard, and OneCLI gateway ports.
- Adds short agent context for diagnosing sandbox network-policy `502 Bad Gateway` errors.
- Updates the README for the current first-run, setup, and customization flow.

## Why

The goal is to keep the Docker kit small and declarative while moving the heavy
work into the published NanoClaw image. Users should not wait for NanoClaw to be
cloned, installed, and built during `sbx run`; the sandbox should start from a
ready host image, apply the network/port policy, and enter NanoClaw setup.

This PR also uses `caps.network.allow` rather than the older
`network.allowedDomains` shape from the reference PR because that is the current
sandbox kit schema validated by `sbx kit validate`.

## Validation

- `sbx kit validate ./nanoclaw`


nanoclaw/spec.yaml
```diff
@@ -1,120 +1,88 @@
 schemaVersion: "1"
```
mdelapenya:
```suggestion
schemaVersion: "2"
```

mdelapenya: Can you execute the migration script to verify all fields?

https://github.com/docker/sbx-kits-contrib/tree/main/scripts#migrate-v1-to-v2go--v1--v2-specyaml-migration

mdelapenya (dismissed):
LGTM! Could you take a look at my comments?

We cannot merge it yet because this kit will land with a v2 format, but the sbx binaries are not supporting v2 spec kits yet (hopefully next release).

But I'm approving anyway to merge it as soon as that happens.

Cheers!

amit-shafnir:
Applied the schema change you requested.
Migration script output: "no changes needed in nanoclaw/spec.yaml"

mdelapenya (approved):

PR #110, opened 2026-06-23
feat(az-devops): new kit - Azure DevOps

tfenster:
**Summary**
Adds a new kit for the Azure DevOps MCP Server and adds authentication via a sbx secret

**Spec choices worth flagging for review**
Narrow selection of allowed domains, but whitelist for *.visualstudio.com as it doesn't work otherwise for older Azure DevOps organizations still using that URL format. This only works with current nightlies, expected to work with sbx 0.34 and later

**Test plan** and **Origin**
See https://tobiasfenster.io/using-docker-sandbox-kits

### DIFF
--- a/az-devops/README.md
+++ b/az-devops/README.md
@@ -1,0 +1,86 @@
+# az-devops
+
+A mixin that enables Azure DevOps API access from a sandbox by wiring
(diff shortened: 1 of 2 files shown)


tfenster:
`sbx kit validate` returns warnings about spec v1 but as nothing here seems to use v2, I assume that is still OK?

PR #118, opened 2026-06-27
Add microsoft-agent-framework kit

shivaylamba:
Adds a v2 mixin that installs the Microsoft Agent Framework (Python) + DevUI into 
  an isolated venv, allows the OpenAI/Azure OpenAI egress domains, and wires the Azure OpenAI 
  credential. OpenAI uses the platform's built-in credential to avoid a compose conflict. 
  Validated end-to-end on sbx v0.33.0 (TCK passes; live OpenAI call through the proxy 
  succeeds). Uses schemaVersion 2 per the repo's kit-author skill.

### DIFF
--- a/microsoft-agent-framework/README.md
+++ b/microsoft-agent-framework/README.md
@@ -18,12 +18,14 @@ First, store your provider key on the host (it never enters the sandbox):
 # OpenAI — uses the platform's built-in `openai` service
 $ sbx secret set -g openai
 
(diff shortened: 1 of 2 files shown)


shivaylamba:
Note: OpenAI is verified end-to-end (live call through the proxy succeeds); for Azure I couldn't fully verify the env-sentinel injection on released sbx v0.33.0,  the spec follows the documented proxyManaged contract, so it should resolve as the engine catches up to the v2 surface (same version where licenses: isn't accepted yet).

PR #100, opened 2026-06-18
feat(kits): migrate all kits to schemaVersion 2 (comment-preserving)

bendebury:
## ⚠️ DRAFT — DO NOT MERGE until a v2-capable sbx release ships

A `schemaVersion: "2"` kit is **rejected by engines that haven't adopted v2**. This PR is parked behind the binding/consent + built-in-v2 migration; un-draft it once a v2-capable sbx release is out.

## What

Migrates **all 20 contrib kits** from `schemaVersion: "1"` → `"2"`.
- 19 kits via `scripts/migrate-v1-to-v2` (`kind: agent`→`sandbox`, `agent:`→`sandbox:`, `network.*`→`caps.network`/`credentials[]`, `environment.proxyManaged`→`credentials[].apiKey`, `memory`→`agentContext`).
- `claude-model-runner` — a clean mixin with no v1-only constructs — bumped **surgically**, because the script is change-driven and leaves such kits at v1.

## Comment preservation

The migration script re-emits through the canonical YAML marshaller, which **drops authors' YAML comments** — egress-domain rationale (why each host is in `caps.network.allow`), pinned-install/supply-chain justifications, filesystem-trust notes, etc. These are high-value docs, so they were **restored** to each migrated spec (137 comment lines across 10 kits).

`#` content inside string values (shell scripts in `commands.*`, markdown in `agentContext`) is preserved by the marshaller and was left untouched.

## Verification

Every kit was checked to:
1. **load + validate as v2** (`LoadFromDirectory`), and
2. parse to a **byte-identical `Artifact`** vs the pre-restore script output — proving the restored comments changed *nothing but comments*.

Full module (`spec`/`bindings`/`scripts`/`tck`) builds and tests green.

## Known migration-script limitations surfaced here (follow-up)

1. **Change-driven:** a clean-v1 kit (no v1-only constructs) isn't bumped to v2 (hit `claude-model-runner`).
2. **Drops YAML comments:** the decode→re-marshal loses comments.

A future fix would bump on `schemaVersion` alone via a surgical edit that preserves comments/anchors. Tracking separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #136, opened 2026-07-09
feat(spec): fork decoder on schemaVersion for clean v2 grammar

cdupuis:
## Summary

Introduce a **schemaVersion-forked loader** in `spec/`. A `spec.yaml` with
`schemaVersion: "2"` is decoded by a new, strict v2 grammar (`spec/v2.go`);
every other value keeps the frozen v1 decoder + `normalize`, so **v1 kits and
their deprecation warnings are unchanged**. Both paths produce the same
canonical `Artifact` and preserve `Manifest.SchemaVersion` — the signal the
credential-binding regime keys on.

The v2 grammar cleans up these field shapes:

| v1 (still accepted) | v2 |
| --- | --- |
| `aiFilename` / `agentContext` | `agentInstructions.filename` / `agentInstructions.content` |
| `commands` / `commands.initFiles` | `setup` / `setup.files` |
| `sandbox.entrypoint.{run,args,ttyArgs}` | flat `sandbox.entrypoint` array + `sandbox.command.{default,interactive}` |
| `sandbox.resources.memoryMB` (int) | `sandbox.resources.memory` (unit string) |
| top-level `caps.network` | `permissions.network.{allow,deny}` |
| `network.publishedPorts` / top-level `publishedPorts` | top-level `ports` |
| `credentials.inject.format` | `credentials.inject.scheme: basic\|bearer` |

`scheme` is decode-time sugar only — it expands to `Format`/`Username` on the
canonical `Artifact`, so downstream consumers are unchanged. The `ports` rename
is likewise a v2-grammar-only key change: the canonical `Artifact.PublishedPorts`
field, its validation, and the v1 spellings are untouched.

Also updates `migrate-v1-to-v2` (plus its golden fixtures) and the v2 test
fixtures to the new grammar.

## Docs

- Adds `spec/SPEC-v2.md` — a formal v2 `spec.yaml` specification with separate
  `kind: sandbox` and `kind: mixin` sections, covering every top-level and
  block-level field, validation rules, composition, and the relationship to v1.
- Rewrites the `kit-author` skill (SKILL.md + topics) to the current v2 grammar
  (`agentInstructions`, `permissions.network`, `setup`, the `entrypoint`+`command`
  split, `resources.memory` byte strings, the `scheme` sugar, and top-level `ports`).
- Adds `spec/OCI-v2.md` — the wire-format spec for a v2 kit as an OCI artifact:
  manifest shape + `artifactType`, media types, the spec-in-config-blob, the
  tar+gzip layer layout (`home/agent/` + `workspace/` prefixes, entry
  constraints, reproducibility pinning), annotations, and producer/consumer
  rules. Cross-linked from `SPEC-v2.md`.

- **The fork keys on `schemaVersion`, not on grammar detection.** `"2"` == new
  grammar, full stop; there is no dual-grammar acceptance for `"2"`.
- **`credentials.inject.scheme` never survives to the canonical `Artifact`** —
  `normalizeV2` expands it into `Format`/`Username` and clears it, and it is
  mutually exclusive with a raw `Format` (validated at load).
- **`sandbox.command.interactive`** enables a TTY session to launch with
  different args than the detached default; when unset it falls back to
  `command.default`. The runtime wiring for this lives in the `sandboxes` repo.

## Why the two catalog kits stay on v1 in this PR

`nanoclaw` and `claude-sbx-statusline` were the only two `schemaVersion: "2"`
catalog kits. Making `"2"` mean the new grammar is a **breaking change** for
them, and `test-kit-e2e` installs the *latest released* `sbx` — whose parser
predates this fork and rejects the new grammar (`field allow not found in
NetworkPolicy`, `field instructions not found in sandboxBlock`). The new
grammar only parses with the parser this PR introduces, which is not in a
release yet.

Both kits only ever used v1-shaped fields, so this PR sets their
`schemaVersion` to `"1"`: the released parser and the new fork's v1 path both
accept them (with the usual v1 deprecation warnings), keeping CI green.
**They will move to the v2 grammar in a follow-up PR** that lands in lockstep
with the `sbx` release carrying this parser.

## Origin

Engine-side change requested to clean up the kit v2 spec grammar; no new kit.

## Test plan

- [x] `go test ./...` green (module-wide).
- [x] `gofmt` + `go vet ./spec/... ./scripts/...` clean.
- [x] `migrate-v1-to-v2` golden fixtures regenerated to the new grammar.
- [x] Both downgraded kits load through the parser without error.

> [!NOTE]
> This PR also removes the stale `TestNanoclawSettingsLift`. NanoClaw stopped
> lifting `~/.claude/settings.json` via `commands.install` in #109 (it manages
> credentials through OneCLI inside its image), but the test was left behind.
> CI never flagged it because `detect-changes` skips `test-tck` for
> nanoclaw-only changes; this PR touches `spec/`, so the job runs again.

> [!IMPORTANT]
> The consuming `sandboxes` repo currently pins this via a temporary local
> `replace` directive. It needs this branch tagged/released and the `replace`
> dropped before that side can merge.


rcjsuen (dismissed):

bendebury:
From our Slack conversation:

Removing the `caps/capabilities/policies` parent block and moving network to the top level--I don't think we should do that, since if we do add future policies then we have to keep adding top level blocks. I get that it seems unnecessarily nested right now, but the intent was to try to be forward-looking.



With regards to the sandbox owning the agent block. We have two types of kits, mixins and sandbox kits. A mixin cannot have a sandbox block. The sandbox is where we define how to run the agent (through the entrypoint ) so it makes sense for other agent-specific configuration to go there. The naming of the profile file is something that is attached to the agent--mixins don't overwrite it. However, mixins can add instructions to that file.

So that means that there is one field that is set ONLY by sandboxes (current aiFilename) and one field which may be populated by both agents and mixins (current agentContext). If you pull out the agent block from the sandbox block, then:
- you break the conceptual idea of "the sandbox is what sets up the agent"
- you have to allow mixins to define an agent block (which I guess I could see but it kinda muddies the split even further than it's currently muddied)
- but you have to ignore a mixin setting the aiFilename (or not allow it) which is awkward

But I see the point about the awkwardness of the split between these two fields which are related to each other.

cdupuis:
Thanks @bendebury. As discussed on Slack I've updated the PR with the changes we agreed. Let me know what you think, please.

bendebury:
LGTM!

bendebury (dismissed):

kiview (dismissed):

bendebury:
Heads-up on cross-PR coordination (cc @cdupuis):

**Merge order:** this PR should land **before** the kit-migration PR #100. #100 had migrated all 23 kits to the *pre-grammar* v2 shape (`commands`/`caps.network`/`agentContext`/`aiFilename`/`publishedPorts`/`inject.format`); once your strict v2 decoder forks on `schemaVersion`, those kits would be `schemaVersion: "2"` in a grammar it no longer accepts. I've put #100 back to draft — the plan is to regenerate it with *your* updated `migrate-v1-to-v2` (new grammar) after this merges. No changes needed on your side; just flagging so #100 doesn't land out from under you.

**One thing worth a look:** the migrator's emit shape still doesn't carry `requires` (base-agent affinity) — see #152. Your rewrite of `migrate-v1-to-v2.go` touches the surrounding code but the diff doesn't add a `requires` field to the output, so a mixin with `requires: agent:` run back through the tool would still drop its affinity. Might be worth folding the fix into this PR while you're in there, or I can follow up on top once this lands.

kiview (dismissed):

doringeman (approved):

PR #138, opened 2026-07-09
feat(playwright): add Playwright browser-automation mixin kit

Iamrushabhshahh:
## Summary

Adds a `playwright` mixin kit that gives any sandbox agent a working browser-automation toolchain: the `playwright` CLI and `@playwright/test` v1.61.1 from npm (version pinned), plus Chromium and its headless shell with all the apt system libraries they need.

With this kit an agent can run end-to-end tests, screenshot pages, generate PDFs, and drive a headless browser against apps served inside the sandbox — with zero project setup.

- **Browsers install to `/opt/ms-playwright`, not the default `~/.cache`.** Install hooks run as root but the agent runs as uid 1000, so a default install would strand the browsers in `/root/.cache` where the agent can't use them. The kit sets `PLAYWRIGHT_BROWSERS_PATH=/opt/ms-playwright` container-wide and leaves the directory world-writable — the same approach as the official Playwright Docker image — so the agent can also fetch a matching browser build at runtime if a project pins a different Playwright version.
- **`NODE_PATH` points at the global npm root.** Node resolves `require('@playwright/test')` by walking up from the test file, so a bare `smoke.spec.js` with no `package.json` can't see the global install. `NODE_PATH=/usr/local/share/npm-global/lib/node_modules` makes zero-config runs work; projects with their own `node_modules` still win, and on templates with a different npm prefix the path simply doesn't exist and Node ignores it.
- **Chromium only.** Keeps the download and the apt footprint small; Firefox/WebKit binaries can be fetched at runtime (the CDN domains are allowed) but their system libraries aren't installed. The README explains how to extend the kit for cross-engine runs.
- **Headless only, by design.** The sandbox has no display server, so `--headed`, `--ui`, and `codegen` won't work. Screenshots, PDFs, traces, and videos all work headless.
- **Version pinning relies on npm's own integrity check.** Unlike the binary-download kits (k8s-tools, trivy, …) there are no manual SHA256s: pinning the exact npm version pins the content, because npm verifies tarballs against the sha512 in the registry metadata. Browser builds are keyed to the Playwright version, so they're transitively pinned.
- **Both browser CDNs are allowlisted — and both are real.** `sbx policy log` from the e2e run shows hits on `cdn.playwright.dev` *and* the `playwright.download.prss.microsoft.com` fallback, so neither entry is speculative.

## Origin

Written for this repository from scratch, following the k8s-tools / trivy kit conventions and the official Playwright Docker image's browser-path approach.

## Test plan

CI runs `kit validate` and the TCK on every PR. CI does **not** run e2e

- [x] `sbx kit validate ./playwright/` passes
- [x] `./scripts/test-kit.sh playwright` passes (the TCK, 100.6s — both install
      commands executed in a real container)
- [x] `./scripts/test-kit-e2e.sh playwright` passes (158s under the `deny-all`
      baseline, scoped to `--app-name sbx-kits-contrib-tck`). Zero bloc
      requests in `sbx policy log`; every `network.allow` entry was observed
      in the log, not guessed.
- [x] Manual smoke: `sbx run --kit ./playwright/ claude` on the scoped
      deny-all daemon, then as the agent user (uid 1000):
      - `playwright --version` → 1.61.1, `PLAYWRIGHT_BROWSERS_PATH` set,
        browsers present in `/opt/ms-playwright`
      - `playwright screenshot "data:text/html,<h1>…</h1>" smoke.png` →
        headless Chromium launched and produced a 9.6 KB PNG
      - a bare `smoke.spec.js` with no `package.json` → `npx playwright test`
        → **1 passed**, resolving `@playwright/test` via the kit's `NOD`

### DIFF
--- a/playwright/README.md
+++ b/playwright/README.md
@@ -1,0 +1,76 @@
+# playwright
+
+A mixin kit that installs the **Playwright** browser-automation toolchain inside the sandbox: the `playwright` CLI and `@playwright/test` **v1.61.1** from npm, plus **Chromium** (and its headless shell) with all required system libraries. Pair it with any agent (Claude, Gemini, …) to let the agent write and run end-to-end tests, scrape or screenshot pages, generate PDFs, and debug web apps served inside the sandbox — all headless, all sandbox-local.
+
+## Usage
+
+```console
+$ sbx run claude --kit ./playwright/ .
+```
+
+Or straight from this repository:
+
+```console
+$ sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=playwright" claude
+```
+
+Prerequisites:
+
+- A base image with Node.js ≥ 18 and npm — all standard agent templates ship it. The install fails loudly with a clear message if npm is missing.
+
+Inside the sandbox:
+
+```console
+$ playwright --version                       # CLI on PATH for every user
+$ npx playwright test                        # run a project's test suite
+$ npx playwright screenshot http://localhost:3000 page.png
+```
+
+Driving a browser from code works with the globally installed library or a project-local one:
+
+```js
+const { chromium } = require('playwright');
+const browser = await chromium.launch();     // headless; finds browsers via PLAYWRIGHT_BROWSERS_PATH
+```
+
+## How it works
+
+### Why browsers live in /opt/ms-playwright
+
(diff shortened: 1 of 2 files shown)


doringeman (approved):
Nice! I tested it and it worked for me!

doringeman:
Merging this to not hold it but be aware of https://github.com/docker/sbx-kits-contrib/pull/136. 🙌 
Thanks for your contribution!

PR #157, opened 2026-07-20
Add neovim kit

manzanit0:
## What
- Add a `neovim` mixin kit that installs latest-stable Neovim (amd64 + arm64) into a sandbox and injects a bundled `~/.config/nvim`.
- Include a ships-a-starter `init.lua` and a `sync-config.sh` helper for bringing your own config.

## Why
- Lets users get a ready-to-use Neovim editor in any agent or shell sandbox without manual setup, and easily carry over their personal config.


dvdksn (approved):
Nice!

PR #162, opened 2026-07-27
Fix code server docs to use 127.0.0.1.

cmrigney:
## Summary

Updated code-server docs to say `127.0.0.1` instead of `localhost` since browsing to `localhost` results in a connection reset.

## Test plan

No test since it's just a doc update.

### DIFF
--- a/code-server/README.md
+++ b/code-server/README.md
@@ -24,7 +24,7 @@ Once the sandbox is up, find the assigned host port:
 $ sbx ports <sandbox-name>
 ```
 
-Open `http://localhost:<host-port>/` in a browser. code-server opens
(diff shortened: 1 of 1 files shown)


robmry:
If the server's only listening on IPv4, maybe it'd be better to change the kit to `protocol: tcp4` (instead of `tcp`). Then "localhost" will work.

robmry:
> If the server's only listening on IPv4, maybe it'd be better to change the kit to `protocol: tcp4` (instead of `tcp`). Then "localhost" will work.

@cmrigney - thank you for letting me know that doesn't work! The tcp4/tcp6 protocol names are missing from the kit spec. This PR fixes the spec, if it's approved and merged an sbx update will still be needed to make it usable.

- https://github.com/docker/sbx-kits-contrib/pull/163

PR #170, opened 2026-08-05
fix(spec): derive v1 deprecation field paths from the v2 grammar

mdelapenya:
## Summary

A v1 kit's deprecation warnings told authors to write field names that **are not valid kit-spec v2 YAML at all**. `network.allowedDomains` pointed at `caps.network.allow` — an internal Go field name; the grammar has used `permissions.network.allow` since the `caps` → `permissions` rename. Anyone following that advice in a `schemaVersion: "2"` spec hits a hard decode error, because the v2 decoder is strict.

Root cause: the paths were hand-typed string literals, so they silently went stale when the grammar moved.

## What changed

- **`spec/fieldpath.go`** — `v1Field()` / `v2Field()` read the dotted YAML path directly off the v1/v2 struct tags via reflection. Every hand-typed field path in `normalize.go`'s `deprecate()` / `notImplemented()` calls now goes through them, so a rename breaks the build and a retag breaks the tests instead of shipping stale guidance.
- **Corrected messages**: `caps.network.allow/deny` → `permissions.network.allow/deny`; `memory` now distinguishes `agentContext` (v1) from `agentInstructions.content` (v2); `settings` distinguishes `commands.startup` (v1) from `setup.startup` (v2); `environment.proxyManaged` notes `apiKey.proxyManaged: true` is also required.
- **`caps:` now warns.** Every other legacy surface warns when used, but a kit writing `caps:` directly got nothing — even though it is itself a superseded spelling. Two kits in this repo (`nanoclaw`, `playwright`) are affected.

One literal remains, with the reason recorded in-code: `credentials.sources` is decoded through a locally-scoped anonymous struct inside `credentialsField.UnmarshalYAML`, which reflection cannot reach off a named field. Its v2 side is still derived.

## Test plan

- `TestFieldPaths` pins all 38 v1/v2 chains used in this package to their documented paths (`spec/SPEC-v2.md`, `skills/kit-author/topics/v1-migration.md`). Verified the alarm actually fires by retagging `permissions` → `grants` and confirming the test fails.
- `go test ./...`, `go vet ./...`, `gofmt` clean.
- **No regressions across real kits**: dumped every warning for all 45 kits in this repo plus the engine's embedded agents, before and after. Every `caps.network.*` became `permissions.network.*`, no warning was lost, and the only additions are the two intended new `caps:` warnings (96 → 98 lines).
- **Migration stays lossless** for the two kits the new `caps:` warning flips from no-op to rewrite: `nanoclaw` and `playwright` produce a byte-identical canonical `Artifact` before vs after `scripts/migrate-v1-to-v2.go`.

## Notes for reviewers

- Independent of #171 (oauth validation) and #172 (V2View) — no file overlap; all three merge cleanly in any order.
- The new `caps:` warning means `migrate-v1-to-v2.go` now rewrites `nanoclaw`/`playwright` instead of treating them as already-clean. That is intended (they should declare `permissions:`), and proven lossless above.

### DIFF
--- a/spec/fieldpath.go
+++ b/spec/fieldpath.go
@@ -7,9 +7,14 @@ import (
 )
 
 // specFileV2Type and specFileType are the reflect roots v2Field and v1Field
-// descend from. Computed once at package init so a typo'd field name in a
-// v1Field/v2Field call panics at program startup (and therefore in any test
-// run), not silently at some later call site.
+// descend from.
+//
+// Resolution is lazy: a bad field name panics when the enclosing v1Field /
+// v2Field call actually runs, not at package init. Since those calls sit
+// inside individual normalize branches, TestFieldPaths (fieldpath_test.go)
+// is what guarantees the drift alarm actually fires — it exercises every
+// path used in this package, so a rename or retag breaks the test suite
+// rather than waiting for a kit that happens to hit that branch.
 var (
 	specFileV2Type = reflect.TypeOf(specFileV2{})
 	specFileType   = reflect.TypeOf(SpecFile{})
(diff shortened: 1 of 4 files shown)


mdelapenya:

```
```

|---|---|---|

cdupuis (approved):

PR #171, opened 2026-08-05
fix(spec): validate credentials[].oauth in ValidateArtifact

mdelapenya:
## Summary

`ValidateArtifact` never validated OAuth credentials at all, so kit authors got no pre-flight signal for a malformed `credentials[].oauth` block. This is the spec-side half (a) of docker/sandboxes#4835.

The only OAuth validator in the package, `ValidateOAuthPolicy`, checks the **pre-normalize v1 `OAuthPolicy` shim** and has **zero callers** — it is exercised only by its own unit tests. The canonical v2 `Credential.OAuth` shape, which both v1 and v2 kits fold onto, had no validator wired in anywhere.

Confirmed reproductions that loaded as `VALID`:

- `credentialFile` with neither `template` nor `structure` → passed validation, then failed at first `sbx create` as an opaque engine error with the real cause buried in `daemon.log`.
- no `sentinels` block → passed validation, and would run with **real OAuth tokens never masked** in the sandbox.
- no `service` on an oauth credential → passed validation, then failed at runtime.

## What changed

- **`ValidateOAuth`** — the v2-shaped analogue of `ValidateOAuthPolicy`, minus `Service` (which v2 moved onto the parent `Credential`). Wired into `ValidateArtifact` over `a.Credentials`.
- **`service` is now required on every credential entry**, per SPEC-v2 §5.4. The engine rejects an empty one at runtime in two places (`NewOAuthInterceptorFromConfig`, and the configure hook's `"oauth service name cannot be empty"`).
- **`SPEC-v2.md` §6** now documents what `ValidateArtifact` checks for credentials — it previously did not mention them at all.
- Drive-by: `lockedPathPattern`'s doc comment used pre-rename example paths (`agent.image`, `network.allowedDomains`).

### Deliberately not enforced

**Sentinels are not required when `passthrough: true`.** SPEC-v2 §5.4.2 states "REQUIRED unless passthrough", and this mirrors the proxy's own runtime guard (`sandboxd/pkg/proxy/oauth_handler.go`) exactly — neither stricter nor looser — so a kit that validates here will not fail the interceptor later. Also required by the existing `v1_compat_test.go` fixture.

**§5.4's lowercase-kebab charset for `service` is not enforced.** The v1 fold derives service keys from env-var names via `deriveServiceKey`, which keeps underscores for any multi-word variable (`SAMPLE_PROXY_TOKEN` → `sample_proxy`, as in `spec/testdata/sample-mixin`). Enforcing it would reject v1 kits that load today. Reason recorded in both the code and §6.

## ⚠️ Behaviour change — worth a minor version bump

This introduces **new hard errors**: a malformed kit that loaded before will now fail `sbx kit validate` and `sbx create`. I verified **none of the 45 kits** in this repo or the engine's embedded agents breaks, and no credential anywhere has an empty `service` — but a third-party kit could. Suggest releasing as **v0.14.0**, not a patch.

## Test plan

- `TestValidateOAuth` (9 subtests) + 4 `ValidateArtifact`-level regression cases tied to #4835, including the passthrough exemption and the non-kebab-service acceptance.
- Verified against the issue's literal repro kits (`neither`, `nosentinel`): both now fail validation with an actionable message instead of reporting `VALID`.
- All 45 kits still load; `go test ./...`, `go vet ./...`, `gofmt` clean.

## Out of scope

- Part (b) of #4835 — placeholder-correctness pre-flight for `credentialFile.structure` (e.g. `{{.AcessToken}}`) — is engine-side, in `docker/sandboxes`.
- `ValidateOAuthPolicy` is left in place as dead-but-exported API rather than removed in this PR.

### DIFF
--- a/spec/SPEC-v2.md
+++ b/spec/SPEC-v2.md
@@ -739,6 +739,19 @@ the per-field rules above:
   byte-size; `mode` octal.
 - **ports**: `container` in 1..65535; `protocol` in `{"", tcp, udp}` (validation error messages spell this `publishedPorts[...]`, the canonical field name).
 - **environment**: variable keys are valid shell identifiers.
+- **credentials**: `service` non-empty on every entry. Its lowercase-kebab
+  charset ([§5.4](#54-credentials)) is **not** enforced here — the v1 fold
+  derives service keys from env-var names and keeps underscores
+  (`SAMPLE_PROXY_TOKEN` → `sample_proxy`), so enforcing it would reject v1
+  kits that load today.
+- **credentials[].oauth** (when present): `tokenEndpoint.host` and
+  `tokenEndpoint.path` non-empty; `sentinels.accessToken` and
+  `sentinels.refreshToken` non-empty **unless** `passthrough: true`;
+  and when `credentialFile` is set, its `path` is non-empty and at least one
+  of `template` / `structure` is present. This mirrors the proxy's own
+  runtime guard, so a kit that validates here will not fail the interceptor
+  later. Placeholder *correctness* inside `structure` is checked by the
+  engine at render time, not here.
 - **setup**: `install[].command` non-empty; `startup[].command` non-empty;
   `files[].path` absolute; `files[].mode` octal; only `${WORKDIR}` placeholder
   in `files[].content`.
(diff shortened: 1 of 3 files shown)


mdelapenya:

```
```

|---|---|---|

cdupuis (approved):

PR #172, opened 2026-08-05
feat(spec): add V2View, the single internal -> v2 grammar projection

mdelapenya:
## Summary

The canonical `Artifact`/`Manifest` field names are the engine's internal vocabulary, and several are **not valid kit-spec v2 YAML keys**: `caps`, `commands.initFiles`, `agentContext`, `template`, `resources.memoryMB`, `publishedPorts`. Anything that renders a kit back out — to disk, or to a human — needs that mapping, and two consumers had each grown their own copy:

- `scripts/migrate-v1-to-v2.go`'s local emit structs
- the engine's `sbx kit inspect --json` (in flight, `docker/sandboxes`)

`spec.NewV2View` becomes the single implementation, carrying **both `json` and `yaml` tags** so the migrator (YAML) and `inspect` (JSON) cannot drift.

Scope is the grammar only — the fields a `spec.yaml` can contain. `Artifact.Warnings` and `Artifact.Files` are not spec keys, so consumers that report them embed the view and add their own fields.

## 🐛 Fixes silent data loss

Found while consolidating: the migrator's hand-maintained emit struct had **no `licenses` or `mixins` field**, so a kit declaring either **lost it on migration, with no warning**. Reproduced on `main` — input with `licenses: [MIT]` + `mixins: [some-mixin]` produced output containing neither. The existing re-parse safety net could not catch it, because dropped output still parses fine.

`licenses` is legally meaningful (SPDX identifiers governing the kit, SPEC-v2 §2).

## What stays local to the migrator

Two concerns are deliberately not in the shared view:

1. **Forcing `schemaVersion: "2"`** — a migrated kit is a fully-declared v2 spec. `NewV2View` carries the version through as-is, so it stays a faithful projection rather than claiming a version the input did not have.
2. **Restoring the author's `entrypoint`/`command` split** from the raw source YAML, via `SetSandboxEntrypoint`. The loader folds a v2 entrypoint tail into `RunOptions`/`InteractiveOptions`, so the split cannot be recovered from the canonical model — having the source YAML is what makes the migration lossless. A consumer without it (e.g. `inspect` on an OCI ref) gets exact per-mode argv but not the author's chosen split.

## Test plan

- **Behaviour-preserving**: the `scripts/testdata/v2-expected` golden fixture is **byte-identical** — I kept `V2View`'s field order matching the old emitter specifically so this is provable rather than asserted.
- **Round-trip test**: the emitted YAML loads back through the v2 grammar with **zero deprecation warnings** — a projection leaking an internal name would fail strict decoding.
- Regression tests for the licenses/mixins loss at both levels (`spec` and `scripts`).
- All 45 kits in this repo plus the engine's embedded agents still load; `go test ./...`, `go vet ./...`, `gofmt` clean.
- Net −85 lines of duplicated emit structs.

## Notes for reviewers

- Independent of the other two open spec PRs — no file overlap; all three merge cleanly in any order.
- The second copy of this mapping disappears when `docker/sandboxes` bumps to the release containing this, which is the plan for the `kit inspect` PR there.

### DIFF
--- a/scripts/migrate-v1-to-v2.go
+++ b/scripts/migrate-v1-to-v2.go
@@ -190,172 +190,36 @@ func migrateSpec(data []byte) ([]byte, []string, error) {
 }
 
 // buildV2 assembles the canonical v2 output spec from a normalized Artifact.
-// The credentials/network/publishedPorts/kind values come from the normalized
-// Artifact (where the v1 → v2 consolidation already happened); the sandbox
-// entrypoint comes from srcSandbox (the raw source block) so the v1
-// run/args/ttyArgs split can be re-expressed as the v2 entrypoint + command
-// grammar. The v1 pipeMode has no v2 equivalent and is dropped.
-func buildV2(a *spec.Artifact, srcSandbox *rawSandbox) *outSpec {
-	out := &outSpec{
-		// Always emit the v2 schema version: a migrated kit is a fully-declared
-		// v2 spec. This is reached only when there were v1 constructs to migrate
-		// (callers gate on a non-empty change list), so a clean v2 spec is never
-		// rewritten just to touch this field.
-		SchemaVersion:  "2",
-		Kind:           a.Manifest.Kind,
-		Name:           a.Manifest.Name,
-		Version:        a.Manifest.Version,
-		DisplayName:    a.Manifest.DisplayName,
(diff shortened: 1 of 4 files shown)


mdelapenya:

```
```

|---|---|---|

cdupuis (approved):

PR #176, opened 2026-08-06
feat(docker): add docker mixin kit; start dockerd in nanoclaw for sbx

Narrator:
Kits that need Docker inside the sandbox.

The new `docker` mixin's single startup command starts dockerd idempotently. Success is keyed on the Docker socket answering, so it composes with local sbx's own dockerd startup. It writes proxy config from `HTTP_PROXY` only when `/etc/docker/daemon.json` is absent, and it fails clearly when the image ships no dockerd. The nanoclaw kit embeds the identical entry (SYNC comment; runtime `mixins:` composition is not applied yet).

Test plan: TCK green for both kits. Script exercised in privileged `shell-docker` containers through the `sudo -u '#0'` wrapper: cold start plus `hello-world`, idempotent re-run, missing-dockerd guard, proxy file. Draft until a stage run passes.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nanoclaw/spec.yaml
+++ b/nanoclaw/spec.yaml
@@ -65,6 +65,57 @@ environment:
     NO_PROXY: "127.0.0.1,localhost"
     no_proxy: "127.0.0.1,localhost"
 
+commands:
+  startup:
+    # SYNCed with the docker mixin (../docker/spec.yaml): update both
+    # together. Runtime mixin composition is not applied yet; when
+    # `mixins:` goes live, replace this block with `mixins: [docker]`.
+    # On local sbx the runtime starts dockerd anyway and this is a no-op.
+    # On `sbx --cloud` this block is what starts the daemon the seed
+    # script and nested agent containers need.
+    - command:
+        - sh
+        - -c
+        - |
+          set -eu
+          # Success is keyed on "the Docker socket answers", never on "this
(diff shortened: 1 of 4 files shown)


nanoclaw/spec.yaml
```diff
@@ -65,6 +65,55 @@ environment:
...
+    # SYNCed with the docker mixin (../docker/spec.yaml): update both
+    # together. Runtime mixin composition is not applied yet; when
+    # `mixins:` goes live, replace this block with `mixins: [docker]`.
+    # On local sbx the runtime starts dockerd anyway and this is a no-op.
+    - command:
+        - sh
+        - -c
+        - |
+          set -eu
+          # Success is keyed on "the Docker socket answers", never on "this
```
mdelapenya, line 79: @Narrator I think we can remove this, as we encourage authors to distribute a Dockerfile (or host it themselves) with all the scaffolding baked into the base image.

Since we added in #192 a pointer to the Nanoclaw base image (see https://github.com/docker/sbx-kits-contrib/pull/192/changes#diff-04fe99ce81daa40882b17dc25cd8ae53a6fff333d1afc3522177a27e68499c7aR7), and there are internal conversations with their team to 1) host themselves the image as of now, or 2) host the image here as a Dockerfile, I would suggest removing this from nanoclaw.

PR #183, opened 2026-08-11
feat(ci): publish and release the kiro kit artifact

mdelapenya:
## Summary

Publishes the kiro **kit artifact** — the spec plus `files/`, which is what `--kit oci://…` consumes without going through git. Until now only the base image (`sbx/kiro-image`) was published; the kit itself had no distribution path.

Two triggers, one push path (`scripts/publish-kit.sh`, called by a reusable workflow):

| Event | Publishes |
|---|---|
| push to `main` touching the kit | `sbx/kiro-kit:<sha>-<YYYYMMDD>` + moves `:latest` |
| tag `kiro/v1.0.0` | `sbx/kiro-kit:v1.0.0`, `:latest` untouched |
| pull request | nothing — validates and prints the plan |

Signed keyless via the job's OIDC token, with a SLSA provenance referrer. Publication is opt-in through `PUBLISH_KITS` (default `kiro`), because every kit here is pushable and discovery would publish all 27.

## Three decisions worth reviewing

**A version describes the kit, not the agent.** `kiro/spec.yaml` gains `version: "1.0.0"`, covering the spec, `files/`, network policy and hooks. The image stays on a floating tag: kiro-cli moves on its own schedule, so pinning it would mean shipping stale agents or churning the kit version for reasons the kit didn't cause. `kiro-kit:v1.0.0` is a stable **contract**, not a reproducible environment.

**Releases don't move `:latest`.** It follows `main`, the tip every kit is tested against on every PR. Reversible with one input if you'd rather it mean "newest release".

**One push, then retag.** `oras.PackManifest` stamps `org.opencontainers.image.created`, so re-pushing the same tree yields a different digest — and each digest carries its own signature and provenance. Pushing twice would give the two tags different attestations for one source. The annotation has one-second resolution, so it's *intermittent*, which is worse. `oras tag` gives one digest, one signature, both tags.

## Guards

`sbx kit push` uses its reference verbatim — it derives nothing from the kit name and validates nothing against it. A mistyped argument would land a kit manifest on the image's tag, in the same namespace, and the registry would accept it. So:

- the reference is composed from the kit directory, never accepted as input
- `check-image-ref.sh` rejects a `sandbox.image` pointing at `<kit>-kit`
- `check-release-tag.sh` refuses a tag whose version disagrees with the spec
- publishing refuses to overwrite an existing tag — an error for a release, a reuse-without-re-push for a dated re-run, so a partial publish stays recoverable

## Runnable locally

The logic is in `scripts/`, not in YAML, so it can be exercised without pushing a branch:

```bash
DRY_RUN=1 MOVE_LATEST=true ./scripts/publish-kit.sh kiro abc-20260811
./scripts/check-release-tag.sh kiro/v1.0.0
```

That was not cosmetic: seven defects in this branch were found and fixed after the logic became runnable — including `oras` not being installed on the release path (which would have published, then failed red) and a PR silently skipping kit validation.

## Blocked on

The Hub OIDC ruleset needs to cover `sbx/kiro-kit`, and `oras tag` needs **pull as well as push** — it fetches the manifest before re-PUTting it. Worth probing before merge, since a push-only credential fails *after* the immutable tag is published:

```console
$ sbx kit push kiro docker.io/sbx/kiro-kit:probe
$ oras tag docker.io/sbx/kiro-kit:probe probe-retag
```

Until the secret and scope are in place every run is a dry run, so this is safe to merge first.

### DIFF
--- a/.github/workflows/hub-overview.yml
+++ b/.github/workflows/hub-overview.yml
@@ -4,12 +4,12 @@ name: Sync Docker Hub overviews
 # and the short description (from spec.yaml).
 #
 # Triggered by DOCUMENTATION changes, not by publishing — and also called by
-# build-kit.yml after a publish, so a repository that has just received its
+# build-and-publish-kits.yml after a publish, so a repository that has just received its
 # first push gets a page without waiting for the next docs edit.
 #
 # Why the docs trigger is the primary one: the Hub page is the only user-facing
 # documentation for these repositories, and a README edit is exactly the change
-# build-kit.yml filters OUT of a rebuild (README.md is not an input to the
+# build-and-publish-kits.yml filters OUT of a rebuild (README.md is not an input to the
 # image). Syncing only on publish would leave the page stale by default and
 # correct by accident, and would make fixing a typo require either an unrelated
 # code change or an artifact re-push that mints a new digest for identical bytes.
@@ -42,7 +42,7 @@ on:
     inputs:
       kit:
         description: |
-          Sync only this kit. Used by build-kit.yml after publishing it; the
+          Sync only this kit. Used by build-and-publish-kits.yml after publishing it; the
           docs trigger leaves it empty and syncs the whole allow-list.
         type: string
         default: ""
@@ -54,7 +54,7 @@ env:
   # A pull request validates and reports, but never writes: the Hub page is
   # production metadata for the default branch, and an unmerged branch has no
   # business overwriting it. Checked here rather than in each caller so every
-  # entry point inherits it — build-kit.yml calls this workflow on PRs too, where
+  # entry point inherits it — build-and-publish-kits.yml calls this workflow on PRs too, where
   # the publish itself is already a dry run.
   DRY_RUN: ${{ github.event_name == 'pull_request' }}
 
--- a/.github/workflows/hub-overview.yml
+++ b/.github/workflows/hub-overview.yml
@@ -152,14 +152,22 @@ jobs:
         if: steps.kit-repo.outputs.ready == 'true' && env.DRY_RUN != 'true'
         uses: peter-evans/dockerhub-description@1b9a80c056b620d92cedb9d9b5a223409c68ddfa # v5.0.0
         with:
-          username: ${{ secrets.DOCKERHUB_SBX_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_SBX_TOKEN }}
+          username: ${{ env.HUB_USERNAME }}
+          password: ${{ env.HUB_TOKEN }}
           repository: ${{ steps.meta.outputs.kit-repository }}
           readme-filepath: ./${{ matrix.kit }}/README.md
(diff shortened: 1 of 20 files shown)


.github/workflows/hub-overview.yml
```diff
@@ -0,0 +1,175 @@
...
+        run: |
+          set -euo pipefail
+          if [ -n "${TOKEN}" ] && [ -n "${USERNAME}" ]; then
+            echo "ready=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "ready=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::DOCKERPUBLICBOT_WRITE_PAT / DOCKERPUBLICBOT_USERNAME unavailable here — Hub overviews are not being synced."
+          fi
+        env:
+          TOKEN: ${{ secrets.DOCKERPUBLICBOT_WRITE_PAT }}
```
ilopezluna: The organization-wide `DOCKERPUBLICBOT_WRITE_PAT` is injected into this shell step on same-repository pull requests. `DRY_RUN` skips the later write actions, but it does not protect the secret: PR-controlled workflow code could modify this step and exfiltrate the PAT. Could we ensure PR validation never receives the secret—for example, use a PR-only gate that emits the validation state without referencing the secret, and condition the credential-bearing gate on `env.DRY_RUN != true`?

mdelapenya: Good catch, and the distinction matters: `DRY_RUN` gated the *write*, not the *credential*. On a same-repository pull request the secret was still materialised into that step's environment, and the PR supplies the workflow code that could read it — so skipping a write that was never the attack is no protection.

Fixed in 7bb5b43, taking your suggestion literally: the secret is not *referenced* at all on the PR path. The gate is split into two mutually exclusive steps.

- `Validation only (pull request)` — `if: env.DRY_RUN == 'true'`, sets `ready=true`, mentions no secret
- `Check the credentials are available` — `if: env.DRY_RUN != 'true'`, the only step that reads them

`ready` becomes `steps.gate.outputs.ready || steps.gate-pr.outputs.ready`, so PR runs still validate metadata and probe both Hub repositories; they just never touch a credential.

I audited the rest rather than assuming this was the only case: parsing every step in every workflow, that gate was the sole unguarded secret reference in the new code — the two sync steps and the OIDC exchange already carried `DRY_RUN` guards.

Then 7c23f4c moves every credential out of inline `with:` into a step-level `env:`, so each is one greppable block per step. For `uses:` steps that is consistency rather than mitigation (no shell, no injection), but it keeps the habit right if one later becomes a `run:`.

One I deliberately left: `e2e.yml` passes `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` to an unconditional step. Those are documented read-only pull credentials so the stakes are lower, but it is the same shape — happy to close it in a follow-up if you think it is worth it.

Original: https://github.com/docker/sbx-kits-contrib/pull/183#discussion_r3764570366

ilopezluna (approved):

PR #190, opened 2026-08-12
fix(ci): use a real Hub PAT for sbx login, not the OIDC token

mdelapenya:
## Summary
- `sbx login` exchanges the password against Docker Hub's real credential API and rejects the OIDC-exchanged registry token `docker/login-action` accepts fine — every real run since #189 failed at sign-in with `docker token is invalid`, before the push even ran.
- Uses the org bot's `DOCKERPUBLICBOT_USERNAME`/`DOCKERPUBLICBOT_WRITE_PAT` instead (already granted to this repo), and gates `DRY_RUN` on both being set so a repo missing either degrades to dry-run rather than a guaranteed failure.
- Scopes `PUBLISHING.md`'s "no long-lived credential" claim to the image push, since the kit artifact's signing step is now the one exception.

## Test plan
- [ ] Merge and watch `kit (kiro) / artifact / publish` go green on `main`

### DIFF
--- a/.github/workflows/publish-artifact.yml
+++ b/.github/workflows/publish-artifact.yml
@@ -65,7 +65,9 @@ env:
   DRY_RUN: >-
     ${{ (github.event_name == 'pull_request'
       || github.repository != 'docker/sbx-kits-contrib'
-      || secrets.DOCKERHUB_OIDC_CONNECTIONID == '') && '1' || '' }}
+      || secrets.DOCKERHUB_OIDC_CONNECTIONID == ''
+      || secrets.DOCKERPUBLICBOT_WRITE_PAT == ''
+      || vars.DOCKERPUBLICBOT_USERNAME == '') && '1' || '' }}
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-artifact.yml
```diff
@@ -121,13 +123,26 @@ jobs:
...
+        # This needs a real Hub password/PAT: the OIDC-exchanged token that
+        # satisfies `docker login` above is rejected by `sbx login` ("docker
+        # token is invalid") — the two authenticate differently. Hence the
+        # org bot's static credential here instead of the OIDC token, the
+        # only static secret in this job.
+        #
+        # Defensive trim: GH preserves whitespace pasted into a secret's
+        # value, and sbx's password-stdin reader only strips a trailing
+        # newline. Hub usernames and PATs are alphanumeric plus hyphens, so
+        # stripping all whitespace cannot corrupt a legitimate value.
```
kiview: I would argue against the need for a defensive trim for a value that is controlled by us, but whatever 🤷

kiview (dismissed):

.github/workflows/publish-artifact.yml
mdelapenya: Agreed. Addressed in [c2919d3](https://github.com/docker/sbx-kits-contrib/pull/190/commits/c2919d37fe6bc91298328ed4e445ea0d39d5bf26)

cdupuis (approved):

kiview (approved):

PR #204, opened 2026-08-16
docs(spec): add section 9, the runtime environment kit content may assume

Narrator:
A kit author has no written answer to basic questions. Which user runs my startup script? May my install hook write /etc? Is jq there? Which env vars does the runtime hand me? The 28 shipped kits answer by copying each other. An author who guesses differently breaks on the next runtime or architecture.

New section 9 pins the profile as normative text. It covers the user model, the root write surface, the tool floor, the two supported architectures, the injected env vars, and lifecycle semantics. For injected env the rule is: names are contract, values are opaque, read them and never hardcode them.

No schema changes. Appended after section 8 so existing anchors keep working. Everything stated is what shipped kits already rely on. `go test ./spec/... ./tck/...` passes.


cdupuis (approved):
LGTM

PR #225, opened 2026-08-20
Add glab kit

johnfeeney-docker:
## Summary

Adds a `glab` mixin kit: installs the GitLab CLI (pinned, digest-verified)
and wires personal access token auth for gitlab.com through the sandbox
proxy. GitLab is not one of sbx's built-in sign-on services (GitHub is), so
today `glab` inside a sandbox has no working auth and `sbx secret set -g
gitlab` has nothing to bind to. This kit closes that gap until native
GitLab sign-on lands.

- Credential is injected as `Authorization: Bearer %s` on `gitlab.com` with
  `proxyManaged: true` (amp-kit pattern): the container only ever sees a
  sentinel in `GITLAB_TOKEN`, which is enough for glab to consider itself
  logged in; the proxy swaps in the real token on outbound requests.
- Install is version + per-arch SHA256 pinned from the release's
  checksums.txt, no `curl | sh` (trivy-kit pattern).
- `network.allowedDomains` is exactly `gitlab.com:443`. I initially allowed
  an object-storage host for the release download redirect, then removed it
  and re-ran e2e under deny-all — release tarballs are served from
  gitlab.com directly, so the narrow allowlist is verified, not guessed.
- Git-over-HTTPS push auth is out of scope — empirically confirmed during
  testing: `git push` fails ("could not read Username"; no credential helper
  or SSH key in the sandbox, and the sentinel token is not usable by git),
  while all `glab` / REST API operations authenticate fine. Documented in
  the kit README, along with a docs-level path for self-managed instances
  via `sbx secret set-custom`.

## Origin

Built to unblock GitLab CLI usage in sandboxes, hit while reproducing a
customer-reported gap (GitLab absent from the typed-service list).

## Test plan

- [x] `sbx kit validate ./glab/` passes
- [x] `./scripts/test-kit.sh glab` passes (the TCK, incl. container install execution)
- [x] `./scripts/test-kit-e2e.sh glab` passes under the deny-all baseline on
      the scoped daemon
- [x] Manual smoke: `sbx run --kit ./glab/ claude` with a real PAT stored via
      `sbx secret set gitlab` — `glab auth status` and `glab api user`
      succeed inside the sandbox with the proxy-injected token
- [x] Manual functional test: from inside the sandbox, created repositories
      in a gitlab.com group, populated them via the REST Commits API
      (content verified byte-for-byte by git blob SHA), and deleted them.
      GET/POST/PUT/DELETE all pass through the proxy with the injected
      token — observed failures were GitLab's own permission model (403 at
      Maintainer, resolved at Owner), not the sandbox.

### DIFF
--- a/glab/README.md
+++ b/glab/README.md
@@ -1,0 +1,59 @@
+# glab — GitLab CLI
+
+A mixin kit that installs the [GitLab CLI (`glab`)](https://gitlab.com/gitlab-org/cli) and wires up personal access token (PAT) authentication for gitlab.com through the sandbox proxy. Pairs with any base agent (claude, codex, gemini, …).
+
+GitLab is not one of sbx's built-in auto-sign-on services (GitHub is), so out of the box `glab` inside a sandbox has no credentials and `sbx secret set -g gitlab` alone has no effect — nothing binds it. This kit closes that gap: it declares a `gitlab` credential and injects it into outbound `gitlab.com` requests at the proxy, so the token is stored on the host and never lands inside the sandbox. The container only ever sees a proxy-managed placeholder in `GITLAB_TOKEN`.
+
+## Usage
+
+Store your GitLab PAT (needs the `api` scope) once on the host:
+
+```console
+sbx secret set gitlab
+```
+
+Then create a sandbox with the kit:
+
+```console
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM!

PR #231, opened 2026-08-21
feat(openclaw): sandbox tool calls in a nested Docker container

mdelapenya:
## Summary

A sandbox that runs OpenClaw should not hand the agent's shell the same filesystem the gateway keeps its token and credential state in. OpenClaw can sandbox tool execution itself, and this kit's image is `shell-docker`, so this turns it on:

```json
"sandbox": { "mode": "all", "backend": "docker", "scope": "session", "workspaceAccess": "rw" }
```

Verified nested, in a sandbox created from this kit with no manual steps:

| | user | host |
|---|---|---|
| outer sbx sandbox | `agent` | `auto` |
| tool call | `sandbox` | `ed812ed462ee` |

`docker ps` inside the sandbox shows `openclaw-sandbox:bookworm-slim` running as `openclaw-sbx-agent-main-main-*`, with `/workspace` mounted. That inner daemon has its own image store, so this is not the host's Docker socket and the Docker-out-of-Docker path-mapping caveat in [Sandboxing](https://docs.openclaw.ai/gateway/sandboxing) does not apply.

## What it needed

**The image has to exist in the inner daemon.** OpenClaw will not substitute a base image when `openclaw-sandbox:bookworm-slim` is missing — the bundled one carries `python3` for the write/edit helpers — and the error it raises points at `scripts/sandbox-setup.sh`, a source-checkout script this npm-installed image does not ship. The bootstrap builds it in the background instead.

**Egress for that build**, taken from a real build under `deny-all` rather than guessed:

| host | why |
|---|---|
| `registry-1.docker.io`, `auth.docker.io` | manifest + auth |
| `production.cloudfront.docker.com` | blob CDN |
| `deb.debian.org`, `security.debian.org` | apt for the base image's packages |

The blob host is **cloudfront**, not cloudflare — the Cloudflare guess fails with `403 Forbidden` on the blob fetch after the manifest succeeds, which is a confusing place to land.

**Readiness now waits for the image**, bounded at 180s. Without it the sentinel means "the gateway listens" while every tool call still fails on the missing image; the e2e prompt subtest hit exactly that and still passed, since any output satisfies its assertion. Bounded so a failed build degrades to a working gateway rather than a sandbox that never reports ready.

## Test plan

- [x] `sbx kit validate`, `test-kit.sh` (TCK), `test-kit-e2e.sh` under `deny-all`
- [x] Fresh sandbox, no manual steps: image built ~10s after boot, tool call lands in the nested container
- [x] e2e prompt subtest returns a real model answer where before the readiness gate it returned `Sandbox image not found`

Independent of #229 (credential modes); both touch `spec.yaml` in different places.

### DIFF
--- a/openclaw/files/home/.local/bin/openclaw-gateway-up.sh
+++ b/openclaw/files/home/.local/bin/openclaw-gateway-up.sh
@@ -110,9 +110,51 @@ if ! gateway_ready; then
     done
 fi
 
+# Tool calls run in a nested Docker sandbox (agents.defaults.sandbox in the
+# shipped openclaw.json), which needs `openclaw-sandbox:bookworm-slim` in this
+# sandbox's own Docker daemon. OpenClaw will not substitute a base image when
+# it is missing, and the error it raises points at a source-checkout script
+# this image does not have, so build it here. Started in the background so the
+# gateway does not wait on a Docker Hub pull plus an apt install.
+SANDBOX_IMAGE_REF=openclaw-sandbox:bookworm-slim
+SANDBOX_IMAGE_WANTED=no
+if command -v docker >/dev/null 2>&1 && ! docker image inspect "$SANDBOX_IMAGE_REF" >/dev/null 2>&1; then
+    SANDBOX_IMAGE_WANTED=yes
+    setsid sh -c "docker build -t $SANDBOX_IMAGE_REF - <<SANDBOX_IMAGE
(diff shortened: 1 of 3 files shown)


mdelapenya:
Folding this into #229 instead of carrying two open PRs for the same kit. The commit is now on that branch unchanged; nothing here is lost.

PR #247, opened 2026-08-27
Add bun kit

kristiyan-velkov:
## Summary
- Add a `bun` mixin that installs Bun v1.4.0 from a SHA256-verified GitHub release zip (`bun-linux-x64` / `bun-linux-aarch64`).
- Extract the binary with Python's `zipfile` so the kit does not need `apt-get install unzip` or Ubuntu/Docker apt hosts.
- Keep the allowlist to the one-shot release download; `bun install` registries are opt-in per sandbox.

- **No `curl | bash` from bun.sh.** Version + per-arch SHA256 are pinned from the release `SHASUMS256.txt`.
- Runtime package installs (`registry.npmjs.org`, etc.) are intentionally *not* allowlisted — same posture as keeping Vale/Task install-only. Documented in the README.
- Python unzip instead of apt `unzip` keeps the network contract to three GitHub asset hosts.

## Origin
New runtime mixin, patterned on `vale/` / `task/` (pinned GitHub release).

## Test plan
- [ ] `./scripts/test-kit.sh bun` passes (the TCK)
- [ ] `./scripts/test-kit-e2e.sh bun` passes under deny-all
- [ ] Manual smoke: `sbx run --kit ./bun/ claude` then `bun --version`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bun/spec.yaml
+++ b/bun/spec.yaml
@@ -1,0 +1,74 @@
+schemaVersion: "2"
+kind: mixin
+name: bun
+displayName: Bun
+description: Installs the Bun JavaScript runtime and package manager from a pinned GitHub release.
+agentInstructions:
+  content: |
+    ## Bun
+
+    The `bun` CLI v1.4.0 is on PATH. Use it as a runtime (`bun run`),
+    a package manager (`bun install`), and a test runner (`bun test`).
+    Run `bun --version` to confirm.
+
+    `bun install` reaches the npm registry and often GitHub; those
+    hosts are not in this kit's allowlist (install time only covers
+    the release tarball). Add them on the sandbox if the project
+    needs to fetch packages:
+
+        sbx policy allow network --sandbox <name> "registry.npmjs.org,github.com,objects.githubusercontent.com"
+permissions:
+  network:
+    allow:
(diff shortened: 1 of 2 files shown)


bun/spec.yaml
```diff
@@ -0,0 +1,74 @@
...
+    # Install Bun from a pinned, SHA256-verified GitHub release.
+    # Do NOT use `curl | bash` from bun.sh — that pattern is unsigned
+    # and unpinned. Do NOT query api.github.com for releases/latest
+    # (CI IPs hit the 60 req/hr limit). To bump: change BUN_VERSION
+    # plus the per-arch SHA256 below (from the release SHASUMS256.txt).
+    # Archives are zip files; python3's zipfile is used so we do not
+    # need an apt unzip and the extra Ubuntu/Docker apt hosts.
+    - command: |
+        set -euo pipefail
+        BUN_VERSION=1.4.0
```
mdelapenya: suggestion: I recommend using a kit-arg for the version, defaulting to latest, or main, or this pinned version, so users can customize the install process

kristiyan-velkov: Ok, make sense. I will do the work. Thanks

PR #249, opened 2026-08-28
fix(git-ssh-sign): scope signing to repos with a remote, and diagnose a dead agent

mdelapenya:
## Summary

Two problems with `git-ssh-sign`, both found while debugging a real signing outage in a long-lived sandbox.

**1. Signing policy was machine-wide.** The kit set `commit.gpgSign=true` in `/etc/gitconfig`, so *every* `git commit` anywhere in the sandbox required a live forwarded SSH agent — including throwaway repositories created by test fixtures that have nothing to do with the user's commits. When the forwarded agent went away, unrelated test suites started failing with a signing error.

This PR scopes the policy to repositories that have at least one remote, via `includeIf "hasconfig:remote.*.url:**"`. The signing *machinery* (`gpg.format`, the key command, the allowed-signers file) stays machine-wide, so `git commit -S` still works in any repository — only the automatic-signing default is scoped.

**2. The failure message pointed at the wrong thing.** Git reports a failed key command as a `warning:` and then emits its own `error: user.signingKey needs to be set for ssh signing`. That second line is the one people act on, and it is misleading — this kit deliberately leaves `user.signingKey` unset, because git only runs `gpg.ssh.defaultKeyCommand` when it *is* unset.

The key command now distinguishes a missing relay socket from an agent that is reachable but holds no usable key, quotes `ssh-add`'s own stderr instead of guessing, explicitly pre-empts git's misleading line, and prints the escape hatch (`git -c commit.gpgsign=false commit …`). The old `[ -z "$SSH_AUTH_SOCK" ]` check was dead code: `SSH_AUTH_SOCK` is always set inside a sandbox, so it never fired. It is now `test -S`, which is what actually distinguishes the common case.

The README and `agentInstructions` gain a recovery runbook that leads with `sbx settings get ssh.agentForwardingEnabled`. In the outage that prompted this, that setting was `false` — which presents inside the sandbox as a *broken* agent rather than a disabled feature, and survives any number of key reloads and sandbox restarts. Checking it first would have turned a long debugging session into one command.

- **`hasconfig:remote.*.url:**` needs git ≥ 2.36.** The install probes for support by building a throwaway repo with a remote and checking whether the include actually fires, rather than parsing a version string. On older git it **fails closed** — signing machine-wide, as today — because an unrecognised `includeIf` keyword evaluates false and would otherwise silently disable signing everywhere.
- **The `**` glob is deliberately bare.** `**` is only a cross-slash wildcard as a whole path component, so `git@github.com:**` does *not* match `git@github.com:org/repo.git`. "Has any remote at all" is the only robust pattern here.
- **`user.signingKey` must stay unset**, and the install keeps `--unset-all`-ing it. Setting it to any value — even a placeholder chosen to produce a nicer error — stops git from ever invoking the key command, which breaks dynamic key resolution in the working case too. There is now a comment saying so.
- **Behaviour change to accept:** a genuinely local-only repository (no remote) stops being signed automatically. That is the point, but it is a narrowing, and a test fixture that *does* add a remote is still subject to signing.
- The install clears any machine-wide `commit.gpgSign`/`tag.gpgSign` an earlier version of this kit left behind, so re-installing converges rather than layering.

## Origin

Existing first-party kit; this is a fix, not a new kit. The scoping problem surfaced when a forwarded agent dropped mid-session and broke `git commit` in unrelated Go test fixtures inside the sandbox.

## Test plan

- [x] `./scripts/test-kit.sh git-ssh-sign` passes (the TCK — includes spec validation, `commands_validation`, and in-container `install_execution`). Re-run after rebasing onto `main`.
- [x] `includeIf` scoping verified directly against git 2.53: a repo **with** a remote resolves `commit.gpgSign=true`; a repo **without** one leaves it unset.
- [x] All four key-command states exercised against the real script: missing socket, agent with no identities, relay up but host agent gone, and healthy (signs, `Good "git" signature`).
- [ ] `sbx kit validate ./git-ssh-sign/` — **not run separately**; `sbx` cannot run inside a sandbox. The TCK's validation subtests cover the same spec checks.
- [ ] `./scripts/test-kit-e2e.sh git-ssh-sign` — **not run**, same reason: needs a scoped `sbx` daemon on a host.
- [ ] Manual smoke (`sbx run --kit ./git-ssh-sign/ <agent>`) — **not run**, same reason.

The three unchecked items need a host run before merge. This kit ships no image and no `network.allowedDomains`, so the cross-arch domain gotchas in the template do not apply.

### DIFF
--- a/git-ssh-sign/spec.yaml
+++ b/git-ssh-sign/spec.yaml
@@ -154,6 +154,19 @@ setup:
         #     error: user.signingKey needs to be set for ssh signing
         # Git's line cannot be suppressed, so the message below has to carry the
         # whole diagnosis and explicitly pre-empt it.
+        #
+        # Braced parameter expansion is banned throughout this file. Released
+        # engines validate initFiles content against a placeholder allowlist
+        # holding WORKDIR alone, and reject every other brace-delimited dollar
+        # form -- it cannot tell shell syntax from a kit-spec placeholder, so
+        # a plain default-value expansion fails the kit at "sbx create" time.
+        # Unbraced "$VAR" is the portable spelling; "set +u" below makes it
+        # behave exactly like the default-value forms it replaces, expanding an
+        # unset variable to the empty string instead of aborting, whatever
+        # options the caller's shell had set. Every variable read here is
+        # either assigned first or explicitly tested for emptiness, so nounset
+        # buys this script nothing.
+        set +u
 
         fail() {
             # A leading newline keeps the body off git's "warning: ...failed: "
@@ -169,7 +182,7 @@ setup:
             exit 1
         }
 
-        sock="${SSH_AUTH_SOCK:-}"
+        sock=$SSH_AUTH_SOCK
 
         if [ -z "$sock" ]; then
             fail "SSH_AUTH_SOCK is not set" \
@@ -200,7 +213,10 @@ setup:
                  "Unexpected state; please report it against the git-ssh-sign kit."
         fi
 
-        config_dir="${GIT_SSH_SIGN_CONFIG_DIR:-/home/agent/.config/git}"
+        config_dir=$GIT_SSH_SIGN_CONFIG_DIR
+        if [ -z "$config_dir" ]; then
+            config_dir=/home/agent/.config/git
+        fi
         mkdir -p "$config_dir"
 
         email=$(git config user.email 2>/dev/null || printf '%s' "agent@sandbox.local")
--- a/git-ssh-sign/spec.yaml
+++ b/git-ssh-sign/spec.yaml
@@ -80,7 +80,14 @@ setup:
         # asks for a signature, and keeping it here means `git commit -S` works
         # in any repository in the sandbox.
         git config --system gpg.format ssh
-        git config --system gpg.ssh.defaultKeyCommand /home/agent/.config/git/ssh-signing-key-command
+        # The key command is invoked through /bin/sh rather than executed
+        # directly. Files shipped under a kit's files/home/ tree land mode
+        # 0644 when the kit is installed from its OCI artifact -- per-file
+        # executable bits are not representable in a v2 layer (spec/OCI-v2.md)
+        # -- and git execs this value itself, so a bare path would die with
+        # "cannot exec: Permission denied" before a single one of the script's
+        # own diagnostics could run.
+        git config --system gpg.ssh.defaultKeyCommand '/bin/sh /home/agent/.config/git/ssh-signing-key-command'
         git config --system gpg.ssh.allowedSignersFile /home/agent/.config/git/allowed_signers
 
         # user.signingKey MUST stay unset: git only runs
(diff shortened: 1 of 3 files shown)


git-ssh-sign/spec.yaml
```diff
@@ -7,43 +7,215 @@ agentInstructions:
...
+        if [ ! -S "$sock" ]; then
+            fail "the forwarded agent socket $sock does not exist" \
+                 "The in-sandbox relay is not running. On the HOST, from a shell where 'ssh-add -l' lists your key: 'sbx start \$SANDBOX_NAME'; if the socket still does not appear, 'sbx daemon restart'."
+        fi
+
+        out=$(ssh-add -L 2>&1)
+        case $? in
+            0) ;;
+            2) fail "nothing answered on $sock (ssh-add: $out)" \
+                    "The relay socket is stale. On the HOST: 'sbx start \$SANDBOX_NAME'." ;;
```
dvdksn: Could we avoid classifying exit status 2 as a stale relay here? When `ssh.agentForwardingEnabled=false`, the relay listener can still accept and immediately close the connection because it has no host socket path, and `ssh-add` reports that as exit 2. That is the original outage this PR is meant to diagnose, but this branch skips the setting check. The suggested recovery also uses `sbx start`, which is not a current top-level command, and `$SANDBOX_NAME` is not available in the host shell. The supported container restart cycle is `sbx stop <sandbox-name>` followed by `sbx run --name <sandbox-name>`. Please make the host-side recovery check `ssh.agentForwardingEnabled` first, restart the daemon when that setting changes, and use the supported stop/run cycle where the container relay itself needs restarting. The same `sbx start` instruction should be corrected in `agentInstructions` and the README too.

git-ssh-sign/spec.yaml
```diff
@@ -7,43 +7,215 @@ agentInstructions:
...
+        #
+        # Git runs this only when user.signingKey is unset, and echoes whatever
+        # this script writes to stderr verbatim as
+        #     warning: gpg.ssh.defaultKeyCommand failed: <stderr>
+        # immediately before its own, misleading
+        #     error: user.signingKey needs to be set for ssh signing
+        # Git's line cannot be suppressed, so the message below has to carry the
+        # whole diagnosis and explicitly pre-empt it.
+        #
+        # Braced parameter expansion is banned throughout this file. Released
```
dvdksn: Could we move this script out of inline `setup.files` content and commit it as an executable static home file at `files/home/.config/git/ssh-signing-key-command`? The kit authoring rule recommends `files/home` for static content, reserving `setup.files` for dynamic `${WORKDIR}` substitution or write-once semantics. More importantly here, the released pre-#200 validator only scans inline init/setup-file content, so a static file can use the ordinary `${SSH_AUTH_SOCK:-}` and `${GIT_SSH_SIGN_CONFIG_DIR:-...}` forms without triggering that compatibility bug. That would let us remove this large validator explanation and the `set +u`/manual-default workaround, shrink `spec.yaml`, preserve the executable bit, and make the shell script easier to test independently.

git-ssh-sign/spec.yaml
```diff
@@ -7,43 +7,215 @@ agentInstructions:
...
-    to verify signatures on existing commits.
+    This sandbox signs git commits with your host's SSH key, forwarded over
+    the sandbox's SSH agent relay. Use `git log --show-signature` to verify
+    signatures on existing commits.
+
+    Automatic signing is scoped to repositories that **have at least one
+    remote**. A throwaway repository created by a test fixture
+    (`git init` with no remote) commits unsigned, so an unrelated test
+    suite is never blocked by signing. `git commit -S` still signs
+    anywhere.
```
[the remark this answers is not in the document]
mdelapenya: Good catch, "never" was wrong. Fixed in 26dacd0 — the exemption now reads "only for as long as it has no remote", and a fixture that goes on to `git remote add` is called out explicitly as still signed.

Your other two locations turned out to be different bugs rather than the same one, so thanks for pointing at them: line 204 is the exit-status-2 misclassification David also raised (fixed in 8cd4165), and line 179 was the key command's closing line asserting the agent is unavailable even in the branch where it's reachable but simply holds no key — that now points at the two lines that carry the real diagnosis.

While I was in there I found two more of the same: the git < 2.36 fallback signs machine-wide, so on an old git nothing is exempt at all, and the README claimed scoping covers "everywhere a commit can actually leave the machine" — commits made after `git init` but before the first `git remote add` are pushed unsigned. Both hedged now.

git-ssh-sign/spec.yaml
mdelapenya: All three accepted — thank you, this is exactly the failure the PR was supposed to catch and I'd sent people down the wrong path for it. Fixed in 8cd4165.

`sbx start` was my mistake and it's gone from `spec.yaml`, the README and the key command; it's `sbx stop <name>` then `sbx run --name <name>` everywhere now. On `$SANDBOX_NAME`: rather than just dropping it, the key command now resolves the sandbox's own name and interpolates the value into the message, so the error prints a host command you can paste as-is.

Recovery leads with `ssh.agentForwardingEnabled`, then the host agent's keys, then `sbx daemon restart` if that setting changed, then the stop/run cycle. I put the daemon restart ahead of the container restart rather than after it — if the setting changed and the daemon hasn't restarted, the fresh container just comes back up with the old one.

One correction, and I'd like your read on it. I couldn't reproduce exit 2 from an accept-then-immediately-close listener: OpenSSH's `ssh-add` gives exit **1** there (`communication with agent failed`), and exit 2 only when the connect itself is refused. So rather than re-pin exit 2 to a specific cause, every non-zero state that isn't a plainly empty agent now routes to the same host-side check with the forwarding setting first — which gets the user to the right place whichever code the disabled forwarder actually produces. If you know it lands on 2 in practice I'm happy to tighten the wording.

git-ssh-sign/spec.yaml
mdelapenya: Agreed, this is a much better shape than what I had — done in 35496dd. The script is now `files/home/.config/git/ssh-signing-key-command`, `set +u` and the long validator explanation are gone, and `${SSH_AUTH_SOCK:-}` / `${GIT_SSH_SIGN_CONFIG_DIR:-…}` are back to the ordinary spelling.

I checked the validator claim before leaning on it, since it's the whole reason the workaround existed. Reconstructing the pre-#200 validator from `9731673^` and running it against two candidate kits: inline content fails with `initFiles[0].content contains unsupported placeholder "${SSH_AUTH_SOCK:-}"`, the identical script under `files/home/` loads fine. The scan only ever ran over `InitFiles`; the `files/` loop never looked at content. `openclaw` already ships `${OPENCLAW_STATE_DIR:-…}` this way in a released kit, so it's proven in the wild too.

One part didn't hold, and it's worth knowing about for other kits: the executable bit does **not** survive. `spec/OCI-v2.md` §4.3 says per-file exec bits aren't representable in a v2 layer and everything lands `0644`, and `openclaw/spec.yaml` has a comment about hitting exactly this. I reproduced it — a 0644 key command gives `fatal: cannot exec '…': Permission denied` and the commit dies before any of our diagnostics print. It only bites the OCI install (`docker.io/sbx/git-ssh-sign-kit:latest`, the README's first example); the `git+https://…#dir=` and local-dir paths preserve the mode, which is why it's easy to miss. So the key command is configured as `/bin/sh <path>` now, same trick openclaw uses. Everything else about the move stands and the kit is smaller for it — thanks for pushing on this one.

dvdksn (approved):
Nice 🚀

PR #261, opened 2026-08-31
Add codex kit

mdelapenya:
## Summary

Publishes the `codex` agent as a public `kind: sandbox` kit. Contrib only — no
engine changes. Six files, modelled on `droid/`:

| File | What it does |
| --- | --- |
| `codex/spec.yaml` | the kit: image, entrypoint, `openai` credential (apiKey + oauth), egress allow-list, env, install + startup hooks |
| `codex/Dockerfile` | `docker.io/sbx/codex-image`, built on `docker/sandbox-templates:shell-docker` |
| `codex/README.md` | usage, the three auth modes, MCP wiring, network policy, base image |
| `codex/README.image.md` | Docker Hub overview for the base image |
| `codex/.dockerignore` | keeps kit metadata out of the build context |
| `codex/testdata/tck.yaml` | `extractedFromBuiltin: true` + the `mcp:` block |

The kit dir and `name:` are both exactly `codex`, because `codex-acp` and
`codex-app-server` pin `requires.agent: codex` and affinity matching is by
exact name. Both mixins' TCK suites still pass against this branch.

No workflow edits: image build/publish is discovered from the presence of
`codex/Dockerfile` plus `sandbox.image`.

**Three fixes to entries carried over from the built-in agent's spec.** These
are the deliberate divergences, not accidents:

1. **MCP header table renamed `headers` → `http_headers`** — the one
   correctness fix, and the important one. Codex spells the key
   `http_headers` and silently ignores unknown config keys, so a `headers`
   table **fails open**: the gateway registers, looks healthy, and every tool
   call through it goes out with no `Authorization` at all. Verified against
   codex-cli 0.151.0 in the built image — with `headers`,
   `codex mcp get mcp-gateway` reports `http_headers: -`; with `http_headers`
   it reports `http_headers: Authorization=*****` and `Auth: Bearer token`.
   `headers` is the correct spelling for the JSON-configured agents
   (`copilot`, `kiro`); it does not transliterate to Codex's TOML.

2. **`corepack npm install -g` → plain `npm install -g`** — the built-in
   image drives the install through `corepack npm` and then attempts
   `corepack disable npm`. Both halves are unsound on this base: `corepack
   npm` downloads an unpinned npm on every build which then warns it does not
   support the image's Node (`npm v12.0.2 does not support Node.js
   v22.22.1`), and `corepack disable npm` cannot succeed at all — the layer
   runs as the non-root `agent` user, so unlinking the root-owned
   `/usr/bin/npm` fails `EACCES` and only the `|| echo` fallback keeps the
   build green. It has nothing to remove either, since nothing ever ran
   `corepack enable`. The end state that matters is unchanged and asserted:
   `/usr/bin/npm` is the real apt npm and is first on `PATH`, which is what
   Codex's self-updater needs. Same `codex-cli 0.151.0`, same install path,
   and the build log is now clean.

3. **`mkdir -p "$HOME/.codex"` added to the MCP startup hook** — small
   hardening. The hook previously went straight to `touch "$cfg"`, which is a
   `set -e` failure if the directory is ever missing. `copilot`'s equivalent
   hook already does this.

**Not fixed, flagged instead** — both are behaviour changes I did not want to
make unilaterally inside a port:

- **`apiKey.inject` includes the bare `openai.com` apex** with an
  `Authorization: Bearer` header. The apex serves OpenAI's website, not the
  API; everything observed from a live CLI goes to `api.openai.com`,
  `chatgpt.com`, `auth.openai.com`, `files.openai.com`. While it stands, any
  process in the sandbox that fetches `openai.com` gets the user's real key
  attached by the proxy. Kept because a missing header is a silent 401 on
  whatever path *did* want it, and "no path wants it" is a negative this port
  cannot prove. There is a `TODO` on the entry and a README section on it —
  **please drop it if you can rule the last path out.**
- **`apiKey` does not set `proxyManaged: true`**, unlike most apiKey-based
  kits here. On the paths this kit configures Codex reads its key from
  `auth.json`, where the kit already writes the `proxy-managed` sentinel
  itself — so the env var is not the delivery path. The consequence is still
  that the real key sits in the container env rather than only in the proxy.
  Turning it on looks more correct and is unverified against Codex's own
  key-discovery order.

**Other deviations from the built-in spec, all deliberate:**

- `sandbox.image` is `docker.io/sbx/codex-image:latest`, not the sandbox
  template. `scripts/check-image-ref.sh` hard-fails anything else for a kit
  that ships a `Dockerfile`. Built `FROM docker/sandbox-templates:shell-docker`.
- Egress hosts are listed **without** port qualifiers. The built-in spec pins
  `:443`/`:80`; every other kit here uses bare hosts, and a hardcoded `:80`
  on the apt mirrors breaks `apt-get update` outright the day the base
  image's sources move to https. `codex-app-server` already proves bare apt
  hosts work under `deny-all`.
- `version: "1.0.0"`, `displayName`, `description` added — kit metadata with
  no built-in equivalent; the TCK requires the latter two.
- `BROWSER=xdg-open` kept for parity even though the base image ships no
  `xdg-utils` (verified absent), so the practical effect is that Codex's
  browser-open attempt fails fast and it falls back to printing the URL.
- The Dockerfile install ends with `codex --version`. An install command's
  exit code only says npm exited 0; `--include=optional` is what actually
  decides whether a usable binary landed, so the outcome is asserted rather
  than assumed.
- `testdata/tck.yaml` sets `configPath` but deliberately **omits**
  `transportKey` and `forbiddenKeys`. Those TCK assertions search for
  *quoted JSON* keys (`"url":`), and Codex's config is TOML (`url = "…"`) —
  declaring `transportKey: url` would fail against a correct script, and the
  `forbiddenKeys` searches could only ever trivially pass. A check that
  cannot fail reads as coverage, so it is left out with the reason recorded
  in the file. Worth a follow-up in the TCK if more TOML-configured agents
  land.

**Known ceiling on CI, please read before trusting the green tick:** while the
`codex` built-in still exists, kit registration rejects the name collision, and
the contrib e2e harness **skips** `extractedFromBuiltin` kits. So green CI here
means TCK plus a local-image build — it does **not** exercise the name path or
a real `sbx create`.

## Origin

Extracted from the `codex` built-in agent shipped inside `sbx`, ported to a
standalone v2 contrib kit. Leg 2 of the embedded-agent extraction (`cursor` is
a separate PR). Prose written fresh; no engine internals reproduced.

## Test plan

CI runs `kit validate` and the TCK on every PR. CI does **not** run e2e on

- [ ] `sbx kit validate ./codex/` passes
- [x] `./scripts/test-kit.sh codex` passes (the TCK)
- [ ] `./scripts/test-kit-e2e.sh codex` passes. The script applies the same
      `deny-all` baseline CI uses and scopes everything to its own daemon
      (`--app-name sbx-kits-contrib-tck`), so my main sbx state is untouched.
      Every entry I added to `network.allowedDomains` came from
      `sbx --app-name sbx-kits-contrib-tck policy log <tck-e2e-…>`, not a guess.
- [ ] Manual smoke: `sbx run --kit ./codex/ codex` and verified the kit's
      binary / files / env are inside the running container.

The three unticked boxes are unticked on purpose: **all of them need `sbx`
against a host daemon, and `sbx` does not run inside a sandbox.** They have not
been run, and nothing below should be read as substituting for them — in
particular, the egress allow-list has **not** been validated under `deny-all`,
which is why the speculative `*.githubusercontent.com` host was left out rather
than guessed in.

What *was* verified, from inside the sandbox:

- `./scripts/test-kit.sh codex` — green, including `install_execution` (the
  install hook runs for real, as the `agent` user, in the built image) and
  `mcp_registration`.
- `./scripts/test-kit.sh codex-acp` and `./scripts/test-kit.sh codex-app-server`
  — both green, so neither mixin regressed and both still resolve their
  `requires.agent: codex` affinity target by exact name.
- `go test ./spec/... ./scripts/...` — green.
- `scripts/check-image-ref.sh docker.io/sbx latest` and
  `scripts/check-release-tag.sh codex/v1.0.0` — both pass.
- `docker build` of `codex/Dockerfile` — succeeds, `codex-cli 0.151.0`, binary
  at `/usr/local/share/npm-global/bin/codex` (exactly the path
  `codex-app-server`'s wrapper hardcodes), `/usr/bin/npm` intact at 9.2.0.
- The install hook run in the real image for all four `SBX_CRED_OPENAI_MODE`
  values (`oauth`, `apikey`, `none`, unset): correct branch each time, valid
  TOML in every case, `auth.json` created in the managed modes and removed in
  the others, and re-running after a mode change leaves no stale block behind.
- The MCP hook run in the real image: no-ops with no gateway, registers once
  with one reserved, stays a single table when run again, and
  `codex mcp get mcp-gateway` / `codex mcp list` confirm Codex actually parses
  the result and attaches the bearer.

### DIFF
--- a/codex/Dockerfile
+++ b/codex/Dockerfile
@@ -1,0 +1,94 @@
+# syntax=docker/dockerfile:1
+
+# Base image for the `codex` sandbox kit.
+#
+# One image, no flavour suffix: the kit picks its own image, so the
+# Docker-in-Docker detail never reaches the user. There is no dockerless variant
+# either — a kind:sandbox kit names a single sandbox.image, so a second image
+# would be unreachable without a second kit to consume it.
+#
+# The base is a floating tag, which is why CI also rebuilds on a schedule.
+ARG BASE_IMAGE=docker/sandbox-templates:shell-docker
+FROM ${BASE_IMAGE}
+
+# Re-declared inside the stage: an ARG defined before the first FROM is a global
+# build arg, visible only to FROM lines. Without this, the LABEL below would
+# expand to an empty string.
+ARG BASE_IMAGE
+
+# CODEX_HOME. Created here rather than only in the kit's install hook so the
+# directory exists — and is owned by `agent`, the USER inherited from the base —
+# before anything else in the container start sequence can create it as root.
+# The install hook still does its own `mkdir -p`, which makes this belt and
+# braces rather than the only guarantee.
+RUN mkdir -p /home/agent/.codex
+
+# Codex's npm package is ~370 MB, so the default fetch timeout is not enough on
+# a slow or proxied network: raise it and retry rather than failing the whole
+# image build on one stalled request.
+#
+# --include=optional is load-bearing, not hygiene. The platform-specific native
+# binary (@openai/codex-linux-x64 and friends) ships as an optional dependency,
+# so omitting the flag produces an install that exits 0 and leaves no working
+# binary behind.
+#
+# Plain `npm`, deliberately. The built-in agent's image drives this install
+# through `corepack npm` and then tries to `corepack disable npm`; both halves
+# are worth dropping here:
+#
+#   * `corepack npm` downloads an unpinned npm over the network on every build
+#     and then warns that that npm does not support this image's Node — a
+#     warning today, a hard failure whenever npm decides to enforce it.
+#   * `corepack disable npm` cannot succeed anyway. This layer runs as the
+#     non-root `agent` user inherited from the base, so unlinking the
+#     root-owned /usr/bin/npm fails EACCES; the `|| echo` fallback next to it
+#     is what keeps the build green. It has nothing to remove either, because
+#     nothing in this image ever ran `corepack enable`.
+#
+# The end state that actually matters is unchanged and verified below: the real
+# apt-installed /usr/bin/npm is the npm on PATH at runtime, which is what Codex
+# needs — its self-updater spawns `npm` directly and dies with ENOENT against a
+# corepack shim.
+#
+# `codex --version` last, deliberately: an install command's exit code says
+# only that npm exited 0, not that a usable binary landed. This asserts the
+# outcome instead of trusting the exit code.
+RUN npm install -g @openai/codex@latest \
+      --include=optional \
+      --fetch-timeout=300000 \
+      --fetch-retries=3 \
+      --fetch-retry-mintimeout=60000 \
+ && codex --version \
+ && command -v npm
+
+# Inherited from the base image, but re-declared deliberately so the value is
+# owned here rather than depending on inheritance from an image this repository
+# does not own.
+#
+# This is a *request* to the runtime, not a description of the image: setting it
+# over a base with no Docker engine yields a sandbox started in Docker mode with
+# nothing to run. Since BASE_IMAGE is overridable, CI asserts the engine is
+# really present rather than trusting this label.
+LABEL com.docker.sandboxes.start-docker="true"
+
+# Both of the labels below OVERRIDE values inherited from the base image, which
+# describe the base rather than this image. Overriding is not optional for
+# `flavor`: left inherited it would read "shell-docker", and sbx would report
+# this image's agent as "shell-docker".
+#
+# sbx reads `flavor` and treats it as an agent identifier — it surfaces the value
+# as an image's `Agent` in the API, and separately uses it (with any `-docker`
+# suffix trimmed) to warn when a template looks built for a different agent than
+# the one being run. So it must be the kit's own name: `codex`.
+LABEL com.docker.sandboxes.flavor="codex"
+
+# Informational only — nothing in sbx reads this. Worth setting because the base
+# is a floating tag rebuilt nightly, so this is the one place the produced image
+# records what it was actually built on.
+LABEL com.docker.sandboxes.base="${BASE_IMAGE}"
+
+# Note: `com.docker.sandboxes=templates` also appears on this image. It is
+# inherited from the base and is not set here — nothing reads it, and this image
+# is not part of that template family, so it is left alone rather than asserted.
+
+CMD [ "codex" ]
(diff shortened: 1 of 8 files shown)


mdelapenya:
The three e2e failures are pre-existing, not from this branch. The same commit runs the identical TCK code on four legs and only fails per sbx build channel: `ecc` fails during kit install on the rc and nightly sbx builds (passes on release), a known ecc install issue, and `qemu` fails at startup ("docker not found", DinD not ready) on the nightly build only. Both die before the assertion phase this PR's TCK changes participate in, and neither kit declares the `mcp:`/`args:` blocks the new code interprets. Reruns reproduce them deterministically per channel, so they're not blocking signals for this diff.

codex/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+#
+# The end state that actually matters is unchanged and verified below: the real
+# apt-installed /usr/bin/npm is the npm on PATH at runtime, which is what Codex
+# needs — its self-updater spawns `npm` directly and dies with ENOENT against a
+# corepack shim.
+#
+# `codex --version` last, deliberately: an install command's exit code says
+# only that npm exited 0, not that a usable binary landed. This asserts the
+# outcome instead of trusting the exit code.
+RUN npm install -g @openai/codex@latest \
```
dvdksn: The upstream template now uses the native installer instead of npm - see https://github.com/docker/sandboxes/pull/5911

kiview (dismissed):
+1 on @dvdksn 's comment.

codex/Dockerfile
mdelapenya: Ok, thanks for the pointer, I'll address it right now

dvdksn:
I see some e2e test failures, not sure if they are flakes or real?

mdelapenya:
@dvdksn the ecc failure comes from the kit installing skills in the sandbox shared dir (there is an in-flight PR engine-side), and the qemu one only fails for the nightly leg, **rc and release legs of the very same qemu kit pass every time**.

dvdksn (approved):

PR #269, opened 2026-09-04
Export tck.RunE2EKit so other kit repos can reuse the real-sandbox e2e layer

jedp-docker:
## Summary
- `TestE2EKit`'s real-sandbox e2e logic lived entirely inside `tck/e2e_test.go`, a `_test.go` file — Go never compiles those into what an importing module gets, so nothing outside this repo could reuse the real-sandbox assertions the way `tck.NewSuiteFromDir`/`RunAll` already let importers reuse the container-based TCK layer.
- Moves that logic into an exported `tck.RunE2EKit(t, kitPath, tck.E2EOptions)` in an ordinary `tck/e2e.go`, parameterized by an app-name (previously hardcoded to this repo's own `sbx-kits-contrib-tck`) so a caller can scope every `sbx` invocation to its own daemon/credential store.
- `tck/e2e_test.go`'s `TestE2EKit` becomes a thin wrapper — same behavior, same `KIT_UNDER_TEST` env var contract.
- Documents the new export in `README.md`'s Packages section.
- Motivation: enables other kit repos (e.g. `sbx-kits-internal`) to import this validator wholesale for their own CI instead of duplicating it, so kit validation elsewhere can't drift from this repo's.

## Test plan
- [x] `go build ./...` and `go build -tags=e2e ./...` pass
- [x] `go vet ./...` and `go vet -tags=e2e ./...` pass
- [x] `go test ./spec/... ./tck/...` pass
- [x] Live TCK run against a real kit (`./scripts/test-kit.sh aider`) passes end to end, confirming the refactor didn't change behavior
- [ ] CI: confirm `e2e-release`/`e2e-nightly` still pass on a real kit change

## Known CI failures (pre-existing, unrelated to this change)
This PR touches `tck/`, which triggers `detect-changes`' "run every kit" rule — the first full-repo TCK sweep in a while. Two kits fail, both due to floating external references unrelated to anything in this diff:

- **`aidlc-claude`** — install step does `git clone --branch v2 ...` against `awslabs/aidlc-workflows`, which has renamed its active branch to `main`. Already fixed in #267 (open); once that merges, this failure should disappear here too.
- **`claude-mem`** — install step (`npx -y claude-mem@latest install`) now fails non-interactively with `provider-selection failed during non-interactive-validation: A provider must be explicit when stdin is not interactive.` This looks like a recent upstream `claude-mem@latest` release added a new interactive provider-selection prompt. No fix PR found for this one yet.


mdelapenya (approved):
LGTM, thanks!

PR #271, opened 2026-09-04
Rename glab kit to gitlab, add gitlab-ssh for git-over-SSH

johnfeeney-docker:
## Summary

- Renames the `glab` mixin to `gitlab` (directory + `name:` field) for clarity.
- Adds a new `gitlab-ssh` mixin, mirroring `github-ssh`: pins GitLab.com's published SSH host keys into `known_hosts` so `git clone`/`push`/`pull` over SSH work without an interactive host-key prompt.
- Updates the `gitlab` kit's README/`agentInstructions` to steer git operations at SSH remotes and point at `gitlab-ssh`.

## Why

The `glab` kit only wires up Bearer auth for `glab`/REST API traffic to `gitlab.com`. Git-over-HTTPS push/pull was never covered, and I tried closing that gap by adding a second `credentials[].apiKey.inject` rule with `scheme: basic` (GitLab accepts `oauth2:<PAT>` as HTTP Basic) on the *same* `gitlab.com` domain.

I tested this against a live sandbox and confirmed it does **not** work: the proxy does not disambiguate two inject rules on one domain by which auth scheme the client already sent, so the Basic-auth git request is never rewritten and GitLab rejects the literal sentinel value with `HTTP Basic: Access denied`. Bearer auth for `glab api` continued to work fine throughout.

GitHub avoids this exact collision because its Bearer traffic goes to `api.github.com` while git-over-HTTPS goes to `github.com` — different domains, no collision. GitLab serves both the API and git smart-HTTP from the single `gitlab.com` host, so there's no domain split to fall back on. This looks like a proxy-level gap (no way to disambiguate same-domain inject rules by scheme), not something fixable from the kit side — I'll file a separate feature request for that.

Given HTTPS push isn't fixable here, `gitlab-ssh` makes the documented workaround (SSH remotes via the forwarded host SSH agent) actually usable non-interactively, the same way `github-ssh` does for GitHub.

## Breaking change note for maintainers

`glab` was merged in #225 and is already published as `docker.io/sbx/glab-kit:latest`. Renaming it to `gitlab` means that published tag goes stale rather than updating — happy to add a redirect/alias if there's a preferred mechanism, or to keep the old name if you'd rather avoid the break. Given it's only been live ~2 weeks I judged the clearer name worth it, but deferring to you here.

## Test plan

- [x] `sbx kit validate` — both kits
- [x] TCK (`scripts/test-kit.sh`) — both kits, passing
- [x] e2e under `deny-all` (`scripts/test-kit-e2e.sh`) — both kits, passing
- [x] Manual verification in a live sandbox:
  - `glab auth status` / `glab api user` — Bearer auth unaffected by the rename
  - Basic-auth git-over-HTTPS clone — confirmed still rejected (`HTTP Basic: Access denied`), documenting the proxy limitation
  - `gitlab-ssh`: `known_hosts` populated at create; `ssh -T git@gitlab.com` passes host-key verification (fails cleanly on `publickey`, not on host trust)
  - Full round-trip with a real registered SSH key: clone + commit + push over `git@gitlab.com:...` succeeded

### DIFF
--- a/gitlab-ssh/README.md
+++ b/gitlab-ssh/README.md
@@ -1,0 +1,73 @@
+# gitlab-ssh
+
+A mixin that pre-populates `~/.ssh/known_hosts` with GitLab.com's SSH host
+keys so SSH operations to GitLab work without interactive host verification
+prompts.
+
+Without this kit, SSH connections from a sandbox to GitLab fail because
+there is no TTY available to interactively accept a new host key.
+
+## Why this exists
+
+The [`gitlab`](../gitlab/) kit wires up PAT auth for `glab` and the GitLab
+REST API over `gitlab.com`, but the sandbox proxy cannot rewrite
+git-over-HTTPS Basic auth on that same domain (see the `gitlab` kit's
+README for why). SSH is the supported path for `git push`/`git pull`/`git
+clone` against GitLab from inside a sandbox. This kit removes the one thing
+that otherwise breaks that path non-interactively: host key verification.
+
+## Prerequisites
+
+Your SSH key must be loaded in the agent on the host and registered with
+your GitLab account:
+
+```console
+ssh-add ~/.ssh/id_ed25519
+```
+
+`SSH_AUTH_SOCK` is forwarded into the sandbox automatically.
+
+Start the sandbox with this kit attached, from its published OCI artifact
+on Docker Hub:
+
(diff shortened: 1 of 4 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #272, opened 2026-09-04
Add vibe kit

k33g:
## Summary

Adds `vibe`, a `kind: sandbox` kit for [Mistral Vibe](https://mistral.ai/products/vibe/),
Mistral AI's open source coding agent.

- `vibe/Dockerfile` — `docker/sandbox-templates:shell-docker` + `mistral-vibe`
  installed from PyPI with `uv tool install`, published by the shared pipeline
  as `docker.io/sbx/vibe-image`. `VIBE_VERSION` pins a release at build time.
- `vibe/spec.yaml` — entrypoint `vibe --trust --agent <arg>`, a proxy-managed
  `MISTRAL_API_KEY`, a four-host allow list, and `~/.vibe` on a 1 GB volume.
- `vibe/README.md`, `vibe/README.image.md`, `vibe/testdata/tck.yaml`.

- **`args` is a first for this repo.** `agent` (default `auto-approve`) selects
  Vibe's agent profile and is referenced from `sandbox.entrypoint` as
  `${{ kit.args.agent }}`. No other kit here uses SPEC-v2 §2.1 args yet, so
  this is the first spec whose entrypoint is only correct after substitution.
  Neither automated surface exercises that path — the TCK overrides the
  entrypoint with `sleep infinity`, and the e2e prompt subtest goes through
  `sbx exec` — so it was verified by hand instead (see the test plan).
- **`--agent auto-approve` by default.** Same call the `crush` (`--yolo`) and
  `grok` (`--yolo`) kits make: the sandbox is the security boundary, and a
  confirmation prompt inside it blocks non-interactive use. Narrowed with
  `--kit-arg agent=ask`.
- **`--trust` in the entrypoint**, because the workspace is the user's own
  project mounted by sbx, and the trust prompt has no TTY to answer it on a
  non-interactive start. It is repeated in `testdata/tck.yaml`'s `promptArgs`,
  since the e2e prompt subtest goes through `sbx exec` and bypasses the
  entrypoint.
- **Deliberately narrow `allow` list** — four hosts, each one read out of the
  installed package rather than guessed: `api.mistral.ai`
  (`DEFAULT_MISTRAL_SERVER_URL`), `chat.mistral.ai` (`DEFAULT_VIBE_BASE_URL`),
  `console.mistral.ai` (browser-auth base and admin-managed config, which 401s
  noisily into `~/.vibe/logs/vibe.log` without it), and `*.mistral.services`
  (the experiments/feature-flag service). No PyPI, no apt hosts: the image is
  pre-baked and `VIBE_ENABLE_AUTO_UPDATE=false`, so nothing installs at runtime.
- **The API key is injected into three hosts, not one.** Vibe sends it to
  `chat.mistral.ai` and `console.mistral.ai` as well as the inference API.
- **`~/.vibe` volume** keeps `config.toml`, sessions, logs and custom agents
  across recreations; it is mounted root-owned, hence the `chown` startup hook.

## Origin

Kit authored by @k33g for this repo, from a working local setup for running
Mistral Vibe under Docker Sandboxes.

## Test plan

CI runs `kit validate` and the TCK on every PR. CI does **not** run e2e on fork
PRs (Docker Hub secrets aren't exposed there), so the e2e step below is required
from my side before requesting review.

- [x] Spec validates clean — `go run ./scripts/verify-kit-spec ./vibe`
      (`valid, no warnings`); the repo's stand-in for `sbx kit validate` where
      no `sbx` binary is available.
- [x] `./scripts/test-kit.sh vibe` passes (the TCK) — builds
      `docker.io/sbx/vibe-image:latest` from the branch's Dockerfile and runs
      validation, network/credential/environment policy, and the container
      subtests against it. Vibe 2.25.0 installs and reports its version.
- [x] `./scripts/test-kit-e2e.sh` passes under the `deny-all` baseline, scoped
      to `--app-name sbx-kits-contrib-tck` — `env`, `tmpfs`, `agentContext` and
      `prompt` all green in 27.6s. The `prompt` subtest answers
      `Error: API error from mistral … Invalid API key`, which is the expected
      shape with no credential stored: the sentinel travelled to
      `api.mistral.ai` and Mistral itself replied, so nothing in the allow list
      was missing and no request was blocked by the policy.
- [x] Manual smoke: `sbx run ./vibe/` — the session starts and a real
      `mistral` secret drives it end to end. That run is also what proves the
      arg substitution: Vibe aborts on an unknown profile
      (`Error: Agent 'x' not found.`), so an unsubstituted
      `${{ kit.args.agent }}` could not have reached a working session. The
      non-default `--kit-arg agent=plan` path has not been exercised
      separately.

Reviewers trying this before the kit merges: `sandbox.image` points at
`docker.io/sbx/vibe-image:latest`, which the pipeline only publishes on merge,
so `sbx run` stops at PREPARE IMAGE with a `403`. Build and side-load it first
— `docker build -t docker.io/sbx/vibe-image:latest ./vibe`, `docker save … -o
/tmp/vibe.tar`, `sbx template load /tmp/vibe.tar` — which is exactly what
`scripts/test-kit-e2e.sh` does on its own scoped daemon.

## Note on commit signatures

The commit carries a DCO `Signed-off-by:` trailer but is not cryptographically
signed — it was made in an environment with no signing key. It will be re-signed
before this PR is ready for review.

### DIFF
--- a/vibe/Dockerfile
+++ b/vibe/Dockerfile
@@ -1,0 +1,51 @@
+# syntax=docker/dockerfile:1
+
+# Base image for the `vibe` sandbox kit: the shell-docker template with
+# Mistral Vibe installed from PyPI.
+#
+# Built and pushed as docker.io/sbx/vibe-image by build-and-publish-kits.yml
+# (see ../PUBLISHING.md). The base is a floating tag and Vibe is installed from
+# a `latest` channel, which is why CI also rebuilds on a schedule.
+#
+# Build it locally the way the test scripts do:
+#   docker build -t docker.io/sbx/vibe-image:latest ./vibe
+ARG BASE_IMAGE=docker/sandbox-templates:shell-docker
+FROM ${BASE_IMAGE}
+
+# Re-declared inside the stage: an ARG defined before the first FROM is a global
+# build arg, visible only to FROM lines. Without this the LABEL below would
+# expand to an empty string.
+ARG BASE_IMAGE
+
+# PyPI version of mistral-vibe: "latest", or an exact number such as 2.25.0 to
+# pin the image to a known release.
+ARG VIBE_VERSION=latest
+
+# uv ships with the template at /usr/local/bin/uv, and `uv tool install` puts
+# the executables in ~/.local/bin, which is already on the template's PATH --
+# so `sandbox.entrypoint` can stay the bare name `vibe`.
+USER agent
+RUN set -ex; \
+    if [ "${VIBE_VERSION}" = "latest" ]; then SPEC="mistral-vibe"; else SPEC="mistral-vibe==${VIBE_VERSION}"; fi; \
+    uv tool install "${SPEC}"; \
+    vibe --version
+
+# Requests Docker-in-Docker from the runtime. Inherited from the base image,
+# but re-declared so the value is owned here rather than depending on
+# inheritance from an image this repository does not own.
+LABEL com.docker.sandboxes.start-docker="true"
+
+# sbx reads `flavor` as the image's agent identifier and surfaces it in the API,
+# so it must be the kit's own name -- left inherited it would read
+# "shell-docker" and sbx would report this image's agent as that.
+LABEL com.docker.sandboxes.flavor="vibe"
+
+# Informational only. Worth setting because the base is a floating tag rebuilt
+# nightly: this is the one place the produced image records what it was built on.
+LABEL com.docker.sandboxes.base="${BASE_IMAGE}"
+
+WORKDIR /home/agent
+
+# The kit sets the real entrypoint (spec.yaml); this only applies to a plain
+# `docker run` of the image, outside sbx.
+CMD ["vibe"]
--- a/vibe/README.image.md
+++ b/vibe/README.image.md
@@ -1,0 +1,9 @@
+# sbx/vibe-image
+
+Base image for the Mistral Vibe kit for [Docker Sandboxes](https://docs.docker.com/ai/sandboxes/).
+
+It is built on `docker/sandbox-templates:shell-docker` and installs [Mistral Vibe](https://github.com/mistralai/vibe) from PyPI with `uv tool install`, as the non-root `agent` user. The `VIBE_VERSION` build argument pins a release; the default, `latest`, is what the nightly rebuild tracks. The image requests Docker-in-Docker through the standard sandbox image label.
+
+## Kit
+
+[`docker.io/sbx/vibe-kit`](https://hub.docker.com/r/sbx/vibe-kit)
--- a/vibe/README.image.md
+++ b/vibe/README.image.md
@@ -2,7 +2,7 @@
 
 Base image for the Mistral Vibe kit for [Docker Sandboxes](https://docs.docker.com/ai/sandboxes/).
 
-It is built on `docker/sandbox-templates:shell-docker` and installs [Mistral Vibe](https://github.com/mistralai/vibe) from PyPI with `uv tool install`, as the non-root `agent` user. The `VIBE_VERSION` build argument pins a release; the default, `latest`, is what the nightly rebuild tracks. The image requests Docker-in-Docker through the standard sandbox image label.
+It is built on `docker/sandbox-templates:shell-docker` and installs [Mistral Vibe](https://github.com/mistralai/mistral-vibe) from PyPI with `uv tool install`, as the non-root `agent` user. The `VIBE_VERSION` build argument pins a release; the default, `latest`, is what the nightly rebuild tracks. The image requests Docker-in-Docker through the standard sandbox image label.
 
 ## Kit
 
--- a/vibe/README.md
+++ b/vibe/README.md
@@ -1,0 +1,78 @@
+# Mistral Vibe
+
+A standalone Docker Sandboxes kit for [Mistral Vibe](https://github.com/mistralai/vibe), Mistral AI's open source coding agent. It runs the `vibe` CLI inside a sandbox with the workspace pre-trusted, tool approval pre-granted, and the Mistral API key held by the sandbox proxy rather than by the container.
+
+## Usage
+
+Use the published kit:
+
+```console
+sbx run --kit "docker.io/sbx/vibe-kit:latest" vibe
+```
+
+Or load it directly from this repository:
+
+```console
+sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=vibe" vibe
+```
+
+Or use a local clone:
+
+```console
+sbx run --kit ./vibe/ vibe
+```
+
+## Authentication
+
+Get a key from the [Mistral console](https://console.mistral.ai/api-keys), then store it on the host under the `mistral` service — the name the kit's `credentials` block declares:
+
+```console
+printf '%s' "$MISTRAL_API_KEY" | sbx secret set mistral
+```
+
+Piping the key in keeps it out of your shell history and out of the process table, where passing it as a flag value would put it. `printf` rather than `echo` because `echo` appends a newline to what it pipes; the sibling kits use `echo` without trouble, so `sbx` evidently trims it, but `printf` leaves nothing to trim and no `401` to diagnose.
+
+`sbx secret set mistral` on its own is equally valid — it prompts for the value on a TTY. Either way the secret is stored once on the host; `sbx` also offers to configure the credential on first launch if none is stored.
+
+Then launch:
+
+```console
+sbx run --kit "docker.io/sbx/vibe-kit:latest" vibe
+```
+
+The container only ever sees `MISTRAL_API_KEY` set to a proxy sentinel. The real key is substituted by the proxy on requests to `api.mistral.ai`, `chat.mistral.ai` and `console.mistral.ai`, and on no other host — so a prompt injection that talks the agent into exfiltrating the variable exfiltrates the sentinel.
+
+## Agent profile
+
+Vibe's [agent profile](https://github.com/mistralai/vibe) decides which tool calls need confirmation. The kit starts `auto-approve`, on the same reasoning as the `crush` and `grok` kits: the sandbox is the security boundary, so a confirmation prompt inside it buys little and blocks non-interactive use.
+
+Pick another one at install time:
+
+```console
+sbx run --kit "docker.io/sbx/vibe-kit:latest" --kit-arg agent=plan vibe
+```
+
+The value is any builtin (`ask`, `plan`, `accept-edits`, `auto-approve`) or a custom agent declared in `~/.vibe/agents/NAME.toml`.
+
+## Persistence
+
+`~/.vibe` is a 1 GB volume, so `config.toml`, sessions, logs, custom agents and `.env` survive recreating a sandbox of the same name. The volume is mounted root-owned, which is why a startup command hands it back to the `agent` user before Vibe writes to it.
+
+## Network
+
+The allow list is the four hosts Vibe reaches for, and nothing else:
+
+| Host | Why |
+| --- | --- |
+| `api.mistral.ai` | Inference API. |
+| `chat.mistral.ai` | Vibe's own base URL — sessions and account state. |
+| `console.mistral.ai` | Admin-managed configuration read at startup; also the browser-auth base URL. |
+| `*.mistral.services` | Feature-flag / experiments service queried at startup. |
+
+Anything else your work needs — a package registry, a git host — has to be added to `permissions.network.allow` or allowed on the host with `sbx policy allow network`.
+
+Telemetry and Vibe's self-update are both switched off through `VIBE_ENABLE_TELEMETRY` / `VIBE_ENABLE_AUTO_UPDATE`, so a run is reproducible and needs no egress to PyPI: the version is whatever the image ships.
+
+## Image
+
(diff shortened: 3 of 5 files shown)


k33g:
Hi @kiview  @mdelapenya @ilopezluna could I get a review on this one when you have a moment? 🙏 
It adds vibe, a kind: sandbox kit for Mistral Vibe

pevers:
Hey @k33g and reviewers, we (Mistral) would love to have this support for Vibe. Let me know if we can do anything getting this change in Docker Sandboxes

mdelapenya (commented):
Nice one, Philippe. Built it locally and it comes up clean: `vibe 2.25.0`, runs as `agent`, TCK green. Reading the allow list out of the installed package instead of guessing is exactly the right instinct, and flagging the `args` situation up front in the description saved me a lot of digging.

Two things I'd like fixed before it goes in (the license and a dead upstream link), the rest is take it or leave it. All the suggestions are committable if you want to batch them.

vibe/spec.yaml
```diff
@@ -0,0 +1,120 @@
+# Docker Sandboxes kit for Mistral Vibe.
+# Usage, prerequisites and design notes: see README.md in this directory.
+schemaVersion: "2"
+kind: sandbox
+name: vibe
+version: "0.1.0"
+displayName: Mistral Vibe
+description: Mistral AI's open source coding agent, with its API key injected by the sandbox proxy
+licenses:
+  - MIT
```
mdelapenya: **question (blocking):** should this be `Apache-2.0`?

Vibe is Apache-2.0 upstream (PyPI metadata for `mistral-vibe` 2.25.0, and the repo itself). Only reason I care: `licenses:` turns into an OCI annotation when the kit gets packed, so it travels with the published artifact.

You're the first kit here to set this field at all, so there was nothing to copy from. Happy for you to just drop the block too.

```suggestion
  - Apache-2.0
```

vibe/README.md
```diff
@@ -0,0 +1,78 @@
+# Mistral Vibe
+
+A standalone Docker Sandboxes kit for [Mistral Vibe](https://github.com/mistralai/vibe), Mistral AI's open source coding agent. It runs the `vibe` CLI inside a sandbox with the workspace pre-trusted, tool approval pre-granted, and the Mistral API key held by the sandbox proxy rather than by the container.
```
mdelapenya: **bug (blocking):** `mistralai/vibe` 404s. The repo is `mistralai/mistral-vibe`, which is also what the PyPI package points at.

Same link on line 47 and in `README.image.md`.

```suggestion
A standalone Docker Sandboxes kit for [Mistral Vibe](https://github.com/mistralai/mistral-vibe), Mistral AI's open source coding agent. It runs the `vibe` CLI inside a sandbox with the workspace pre-trusted, tool approval pre-granted, and the Mistral API key held by the sandbox proxy rather than by the container.
```

vibe/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```console
+sbx run --kit "docker.io/sbx/vibe-kit:latest" vibe
+```
+
+The container only ever sees `MISTRAL_API_KEY` set to a proxy sentinel. The real key is substituted by the proxy on requests to `api.mistral.ai`, `chat.mistral.ai` and `console.mistral.ai`, and on no other host — so a prompt injection that talks the agent into exfiltrating the variable exfiltrates the sentinel.
+
+## Agent profile
+
+Vibe's [agent profile](https://github.com/mistralai/vibe) decides which tool calls need confirmation. The kit starts `auto-approve`, on the same reasoning as the `crush` and `grok` kits: the sandbox is the security boundary, so a confirmation prompt inside it buys little and blocks non-interactive use.
```
mdelapenya: **bug (blocking):** same 404.

Upstream has an anchor for this specific bit, so it can point at the profile docs rather than the repo root.

```suggestion
Vibe's [agent profile](https://github.com/mistralai/mistral-vibe#built-in-agents) decides which tool calls need confirmation. The kit starts `auto-approve`, on the same reasoning as the `crush` and `grok` kits: the sandbox is the security boundary, so a confirmation prompt inside it buys little and blocks non-interactive use.
```

vibe/README.image.md
```diff
@@ -0,0 +1,9 @@
+# sbx/vibe-image
+
+Base image for the Mistral Vibe kit for [Docker Sandboxes](https://docs.docker.com/ai/sandboxes/).
+
+It is built on `docker/sandbox-templates:shell-docker` and installs [Mistral Vibe](https://github.com/mistralai/vibe) from PyPI with `uv tool install`, as the non-root `agent` user. The `VIBE_VERSION` build argument pins a release; the default, `latest`, is what the nightly rebuild tracks. The image requests Docker-in-Docker through the standard sandbox image label.
```
mdelapenya: **bug (blocking):** same 404, and this is the one I'd most want fixed.

This file gets synced onto the Docker Hub page for `sbx/vibe-image`, so the dead link ends up there too.

```suggestion
It is built on `docker/sandbox-templates:shell-docker` and installs [Mistral Vibe](https://github.com/mistralai/mistral-vibe) from PyPI with `uv tool install`, as the non-root `agent` user. The `VIBE_VERSION` build argument pins a release; the default, `latest`, is what the nightly rebuild tracks. The image requests Docker-in-Docker through the standard sandbox image label.
```

vibe/spec.yaml
```diff
@@ -0,0 +1,120 @@
...
+  #   trust prompt has nothing left to protect against and would only block a
+  #   non-interactive start.
+  # --agent: the sandbox is the security boundary, so the default profile lets
+  #   Vibe act without per-tool confirmation -- the same call the crush
+  #   (--yolo) and grok (--yolo) kits make. Narrow it with --kit-arg agent=ask.
+  entrypoint:
+    - vibe
+    - --trust
+    - --agent
+    - "${{ kit.args.agent }}"
```
mdelapenya, line 42: **note (non-blocking):** closing the loop on your PR description, nothing to do here.

You're right that neither surface exercises the substitution. For the record, the TCK swaps the entrypoint out for `sleep infinity` and the e2e prompt subtest goes through `sbx exec`, so neither one ever touches it.

That's #278 and I'm on it right now. Yours is the first kit in the repo to declare `args:`, so it's the fixture I'll test the fix against.

vibe/spec.yaml
```diff
@@ -0,0 +1,120 @@
...
+    allow:
+      # Inference API (DEFAULT_MISTRAL_SERVER_URL).
+      - "api.mistral.ai:443"
+      # Vibe's own base URL (DEFAULT_VIBE_BASE_URL) -- sessions and account
+      # state.
+      - "chat.mistral.ai:443"
+      # Console, which is also the browser-auth base URL: Vibe reads the
+      # organization's admin-managed configuration from it at startup. Without
+      # it that call 401s into ~/.vibe/logs/vibe.log -- non-fatal, but noisy.
+      - "console.mistral.ai:443"
```
mdelapenya: **suggestion (non-blocking):** I think these two are swapped.

`chat.mistral.ai` is `DEFAULT_VIBE_BASE_URL` and where the admin-managed config gets fetched from. `console.mistral.ai` is the `/whoami` account and plan lookup. The entries themselves are right, it's just the comments describing each other's host.

Small thing in the same comment: a host missing from `allow` gets blocked by the proxy rather than 401ing. The 401 is what a missing `inject` entry looks like.

```suggestion
      # Vibe's own base URL (DEFAULT_VIBE_BASE_URL), which is also where the
      # organization's admin-managed configuration is read at startup.
      - "chat.mistral.ai:443"
      # Console: the /whoami account and plan lookup, and the browser-auth
      # base URL.
      - "console.mistral.ai:443"
```

vibe/spec.yaml
```diff
@@ -0,0 +1,120 @@
...
+      # Feature-flag / experiments service (experiments.mistral.services),
+      # queried at startup.
+      - "*.mistral.services:443"
```
mdelapenya: **question (non-blocking):** exact host rather than the wildcard?

`experiments.mistral.services` is a constant upstream, so there's no per-tenant subdomain to cover. And it only gets queried when telemetry is on, which you switch off a bit further down, so as configured the call never fires.

Worth noting `mistral.services` is a different registrable domain from `mistral.ai`, so the wildcard is a wider grant than it looks sitting next to the other three.

```suggestion
      # Feature-flag / experiments service. Only reached when telemetry is
      # enabled, which this kit disables below.
      - "experiments.mistral.services:443"
```

vibe/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+USER agent
+RUN set -ex; \
+    if [ "${VIBE_VERSION}" = "latest" ]; then SPEC="mistral-vibe"; else SPEC="mistral-vibe==${VIBE_VERSION}"; fi; \
+    uv tool install "${SPEC}"; \
+    vibe --version
+
+# Requests Docker-in-Docker from the runtime. Inherited from the base image,
+# but re-declared so the value is owned here rather than depending on
+# inheritance from an image this repository does not own.
+LABEL com.docker.sandboxes.start-docker="true"
```
mdelapenya, line 36: **question (non-blocking):** does Vibe actually need Docker-in-Docker?

Asking rather than asserting. Vibe doesn't shell out to `docker` as far as I can see, and with no registry in the allow list the daemon can't pull anything anyway. `docker/sandbox-templates:shell` would be the lighter base.

Counter-argument I'll make on your behalf: a coding agent runs the user's build and test commands, and for a containerised project those are `docker` commands. It's a real trade and I don't think you got it wrong.

Full disclosure, this rule currently only exists inside nine Dockerfile comments in a PR of mine that hasn't landed yet, so there was nowhere for you to read it. I'm moving it into the kit-author docs. Happy to just fold this into my PR and leave yours alone.

vibe/spec.yaml
```diff
@@ -0,0 +1,120 @@
...
+    VIBE_ENABLE_AUTO_UPDATE: "false"
+    # Never block on a git credential prompt in a non-interactive session.
+    GIT_TERMINAL_PROMPT: "0"
+
+setup:
+  startup:
+    # ~/.vibe is a block volume (see `volumes`). It is mounted root-owned, so
+    # it has to be handed back to the agent user on every start, before Vibe
+    # writes to it. Idempotent, as every startup command must be.
+    - command: ["sh", "-c", "mkdir -p /home/agent/.vibe && chown -R agent:agent /home/agent/.vibe"]
```
mdelapenya: **suggestion (non-blocking):** `-R` only does anything on the very first start.

Startup commands run on every container start, not just create. First time round the volume is empty so it's a single chown anyway, and after that everything under there was written by Vibe as `agent` already, so it re-walks a growing pile of sessions and logs to change nothing.

```suggestion
    - command: ["sh", "-c", "mkdir -p /home/agent/.vibe && chown agent:agent /home/agent/.vibe"]
```

vibe/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Authentication
+
+Get a key from the [Mistral console](https://console.mistral.ai/api-keys), then store it on the host under the `mistral` service — the name the kit's `credentials` block declares:
+
+```console
+printf '%s' "$MISTRAL_API_KEY" | sbx secret set mistral
+```
+
+Piping the key in keeps it out of your shell history and out of the process table, where passing it as a flag value would put it. `printf` rather than `echo` because `echo` appends a newline to what it pipes; the sibling kits use `echo` without trouble, so `sbx` evidently trims it, but `printf` leaves nothing to trim and no `401` to diagnose.
```
mdelapenya: **suggestion (non-blocking):** I'd cut this after the first sentence.

The `printf` advice is good. The rest reasons about what sbx does with the trailing newline, and our own docs use `echo "$KEY" | sbx secret set ...` as the worked example, so it ends up reading like the docs are getting away with something.

```suggestion
Piping the key in keeps it out of your shell history and out of the process table, where `-t/--token` would put it.
```

vibe/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```console
+sbx run --kit "docker.io/sbx/vibe-kit:latest" --kit-arg agent=plan vibe
+```
+
+The value is any builtin (`ask`, `plan`, `accept-edits`, `auto-approve`) or a custom agent declared in `~/.vibe/agents/NAME.toml`.
+
+## Persistence
+
+`~/.vibe` is a 1 GB volume, so `config.toml`, sessions, logs, custom agents and `.env` survive recreating a sandbox of the same name. The volume is mounted root-owned, which is why a startup command hands it back to the `agent` user before Vibe writes to it.
```
mdelapenya: **suggestion (non-blocking):** worth one more sentence on `.env`.

Upstream that file is Vibe's own API key store, so anyone who knows Vibe will read this and wonder whether their real key is being persisted onto the volume. It isn't, since the proxy-managed env var wins, but the README is the only place that can say so.

```suggestion
`~/.vibe` is a 1 GB volume, so `config.toml`, sessions, logs, custom agents and `.env` survive recreating a sandbox of the same name. `.env` is Vibe's own key store; the environment takes precedence over it, so the proxy-managed `MISTRAL_API_KEY` is what Vibe uses regardless of what lands there. The volume is mounted root-owned, which is why a startup command hands it back to the `agent` user before Vibe writes to it.
```

vibe/Dockerfile
k33g: ok 👍

kiview (approved):

k33g:
@pevers merged :)

pevers:
> @pevers merged :)

Thank you!

PR #274, opened 2026-09-06
chore(claude-mem): stop setting DO_NOT_TRACK sandbox-wide

winky:
## What

Drops `DO_NOT_TRACK: "1"` from the `claude-mem` kit's `environment.variables`, keeping `CLAUDE_MEM_TELEMETRY: "0"`. Updates the two places that documented the old pair (`agentInstructions` and the README design note).

No change to what claude-mem itself reports.

## Why

**It is redundant.** Upstream resolves telemetry consent through one function (`src/services/telemetry/consent.ts`), used by both the worker and the CLI:

```
1. DO_NOT_TRACK set (truthy) -> always off
2. CLAUDE_MEM_TELEMETRY env: '0'/'false'/'off' -> off, '1'/'true'/'on' -> on
3. telemetry.json config
4. Default: on
```

`CLAUDE_MEM_TELEMETRY=0` alone resolves to off at step 2. The installer's opt-in prompt is also unreachable here for a different reason — it returns early on non-interactive stdin, before it ever reads `DO_NOT_TRACK`.

**It is broader than this mixin's remit.** `DO_NOT_TRACK` is the cross-tool convention from [consoledonottrack.com](https://consoledonottrack.com/). Setting it in `environment.variables` applies it to the whole sandbox, so adding this memory mixin also silences the base `claude` kit and any other tool the user runs inside. That is a sandbox-wide policy decision, and a mixin shouldn't make it on the user's behalf.

**It is inconsistent with the rest of the repo.** `claude-mem` is the only kit here that sets `DO_NOT_TRACK`. The established pattern is vendor-scoped: `paperclip` sets `PAPERCLIP_TELEMETRY_DISABLED: "1"`, and `kiro` allow-lists its telemetry hosts rather than suppressing them.

The durable half of the original intent is untouched: `us.i.posthog.com` is still absent from `permissions.network.allow`. That is the stronger guarantee of the two, since it holds even if upstream renames the variable — which matters for a kit that tracks `@latest` on purpose.

## Verification

Run on this branch:

| | |
|---|---|
| `sbx kit validate ./claude-mem/` | VALID |
| `../scripts/test-kit.sh` | `environment_policy/variables` and `container/environment_variables` PASS |

`container/install_execution` fails on this branch, but for an unrelated pre-existing reason — the install step aborts without `--provider`, which is what #273 fixes.

To confirm the two changes compose, I merged this branch with #273 locally and re-ran everything:

| | |
|---|---|
| `sbx kit validate ./claude-mem/` | VALID |
| `../scripts/test-kit.sh` | PASS in full — `install_execution` 52.1s |
| `../scripts/test-kit-e2e.sh` | PASS under `deny-all` — install completes in 156.2s |

The two branches merge cleanly.

## Note

This depends on nothing in #273 and can land in either order, though `test-kit.sh` won't be fully green on this branch alone until #273 merges.

### DIFF
--- a/claude-mem/README.md
+++ b/claude-mem/README.md
@@ -63,9 +63,14 @@ sbx ports <sandbox> --publish 37700/tcp
   driven by `SBX_CRED_ANTHROPIC_MODE`) and the `enabledPlugins` entry
   are present, never overwriting existing keys. Trace at
   `/tmp/claude-mem-reconcile.log`.
-- **Telemetry off at the source**: upstream's PostHog telemetry is ON by
-  default; the kit sets `DO_NOT_TRACK=1` + `CLAUDE_MEM_TELEMETRY=0` and
-  does not allow-list `us.i.posthog.com`.
+- **Telemetry off at the source, scoped to claude-mem**: upstream's
+  PostHog telemetry is ON by default; the kit sets
+  `CLAUDE_MEM_TELEMETRY=0` and does not allow-list `us.i.posthog.com`.
+  The cross-tool `DO_NOT_TRACK` convention is deliberately *not* set — it
+  would silence the base claude kit and every other tool in the sandbox,
+  which is not a mixin's call to make. The missing allow-list entry is
+  the durable half of this: it holds even if upstream renames the
+  variable.
 - First memory compression uses your existing claude auth (the proxy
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #275, opened 2026-09-07
test(tck): skip e2e when an agent needs a host-stored credential

mdelapenya:
## Summary

The `aidlc-claude` kit has been failing e2e on `main` since it merged ([run 33415633658](https://github.com/docker/sbx-kits-contrib/actions/runs/33415633658)). It declares `requires.agent: claude-bedrock`, and creating that agent needs an AWS profile stored via `sbx secret set bedrock` — which CI does not have, and should not have. `sbx create` reports only a generic `failed to run sandbox container`, so the cause is invisible in the log, and the script's failure epilogue then pointed at the network allowlist, which was a dead end.

Kits can now declare that dependency:

```yaml
# aidlc-claude/testdata/tck.yaml
requiresHostCredentials: ["bedrock"]
```

The harness probes `sbx secret ls --service <svc> --json` before `sbx create` and skips the kit when nothing is stored. It probes for *presence* rather than matching the create failure text, since that text says nothing about credentials and matching it would be fragile. A probe that errors or returns an unrecognized payload still **fails** — only a confirmed-absent credential produces a skip, so this is not a blanket "skip e2e for this kit". Unlike `extractedFromBuiltin`, the flag does not self-obsolete: the requirement is permanent for the agent.

## Follow-ons, each its own commit

- **`fix(aidlc-claude)`** — `awslabs/aidlc-workflows` deleted its `v2` branch, so the kit's install step failed with `Remote branch v2 not found in upstream origin` entirely independently of the above. The harness it copies now lives on `main`. The per-project SHA pin in `aidlc/.aidlc-workflows-version` is unaffected: the clone still resolves a branch tip once and records that commit. This commit is self-contained and reverts cleanly on its own if you would rather handle the kit separately.
- **`fix(scripts)`** — the e2e failure epilogue told authors to widen their network allowlist even when the policy log reported no blocked requests. It now distinguishes three outcomes: blocked rows found, none found, or the log could not be read.
- **`docs`** — contributor docs still named the v1 `network.allowedDomains` field, and the README's new-kit tutorial showed a `schemaVersion: "1"` spec whose grammar is a hard error under v2.

## Why this reached `main`

Both e2e legs are gated on `head.repo.full_name == github.repository`, and the `e2e` job treats `skipped` as a pass. #238 came from a fork, so both legs were skipped and it merged green — the gap this repo's own PR template already warns about. This PR is same-repo, so e2e runs for real.

Note the CI matrix here is large: changes under `tck/**` escalate `detect-changes` to every kit, so all of them run. That is the intended coverage for a shared-harness change.

## Test plan

- [x] `go test ./spec/... ./tck/...` (the `test-tck` job) passes
- [x] `go test ./scripts/...` (the `test-scripts` job) passes
- [x] `./scripts/test-kit.sh aidlc-claude` passes, **including `install_execution`** — this is what the `v2`→`main` fix restores, and it exercises the kit for real (Bun install + harness clone + file placement in the container)
- [ ] `./scripts/test-kit-e2e.sh aidlc-claude` — **not run.** No `bedrock` profile is available to me, so the "credential stored → e2e runs in full" path is unexercised, and CI cannot exercise it by design. The absent-credential path is covered by `TestSecretServiceStored` and will show as a SKIP on this PR's `aidlc-claude` e2e leg. Flagging this rather than ticking the box.

The new `TestSecretServiceStored` deliberately sits in an untagged file so the fast `test-tck` job runs it — the e2e leg invokes `go test -run TestE2EKit`, which would otherwise filter it out and leave it running nowhere.

### DIFF
--- a/aidlc-claude/testdata/tck.yaml
+++ b/aidlc-claude/testdata/tck.yaml
@@ -1,0 +1,8 @@
+# TCK configuration for the aidlc-claude kit.
+#
+# aidlc-claude declares `requires.agent: claude-bedrock` (spec.yaml), and
+# creating a claude-bedrock sandbox fails unless an AWS profile is already
+# stored on the host via `sbx --app-name sbx-kits-contrib-tck secret set
+# bedrock`. CI stores no such per-agent cloud credential, so this kit's e2e
+# is skipped there and must be run locally instead.
+requiresHostCredentials: ["bedrock"]
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,7 +14,7 @@ they run before merge. See the checklist below.
 ## Spec choices worth flagging for review
 
 <!-- Decisions a reviewer should sanity-check: unusual image, deliberately narrow
-allowedDomains, workaround for a known bug, etc. -->
+permissions.network.allow, workaround for a known bug, etc. -->
 
 ## Origin
 
@@ -31,7 +31,7 @@ required from your side before requesting review.
 - [ ] `./scripts/test-kit-e2e.sh <kit>` passes. The script applies the same
       `deny-all` baseline CI uses and scopes everything to its own daemon
       (`--app-name sbx-kits-contrib-tck`), so my main sbx state is untouched.
-      Every entry I added to `network.allowedDomains` came from
+      Every entry I added to `permissions.network.allow` came from
       `sbx --app-name sbx-kits-contrib-tck policy log <tck-e2e-…>`, not a guess.
 - [ ] Manual smoke: `sbx run --kit ./<kit>/ <agent>` and verified the kit's
       binary / files / env are inside the running container.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ For kits that have a corresponding tutorial on [docs.docker.com](https://docs.do
 
 ## Network policy: declare every domain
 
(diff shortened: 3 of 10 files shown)


aidlc-claude/testdata/tck.yaml
```diff
@@ -0,0 +1,8 @@
+# TCK configuration for the aidlc-claude kit.
+#
+# aidlc-claude declares `requires.agent: claude-bedrock` (spec.yaml), and
+# creating a claude-bedrock sandbox fails unless an AWS profile is already
+# stored on the host via `sbx --app-name sbx-kits-contrib-tck secret set
```
dvdksn, line 5: why are we referencing --app-name here (and in many other places)?

dvdksn (dismissed):

aidlc-claude/testdata/tck.yaml
mdelapenya: The app name scopes the TCK to its own daemon so a local run can't disturb your real sandboxes — `CONTRIBUTING.md` and the kit-author docs cover that part.

Why it shows up in *this* file: the line above it declares that this kit's agent can't be created at all without a stored credential, and the probe that checks for it runs against that scoped daemon. So the actionable form of the requirement is a command against that daemon — drop the flag and following this comment would store the secret where the probe never looks, and the e2e would skip anyway.

On the code side it's a single `appName` constant now; that change removed three hardcoded copies that were already in `e2e_test.go`.

kiview (approved):

PR #276, opened 2026-09-07
fix(claude-mem,claude-ollama): enumerate chown paths under ~/.claude

mdelapenya:
Same root cause as the claude-sbx-statusline fix in #275: the shared skills directory under ~/.claude is mounted read-only by default, and a recursive chown over the parent recurses into it and fails.

- claude-ollama: the install step's chown -R hard-fails sandbox create.
- claude-mem: the startup trap's chown -R fails silently, leaving root-owned files under ~/.claude.
- Both now chown exactly the paths their root steps create. SPEC-v2.md's ownership example taught the recursive idiom and is updated to the enumerated one.

Validation:
- TCK passes for claude-ollama (the new chown line executes in-container).
- claude-mem's TCK fails on a pre-existing upstream installer regression, unrelated and reproduced on main: docker/sandboxes#5714.
- The TCK cannot reproduce the read-only skills mount; that behavior is only verifiable e2e against a current engine.

Note: this touches spec/, which expands the e2e matrix to all kits — the ecc nightly leg is expected red (docker/sandboxes#5539).


claude-mem/spec.yaml
```diff
@@ -70,9 +70,19 @@ setup:
...
+    # account.
+    #
+    # Pinned to 13.24.1 instead of `@latest`: an unpinned install started
+    # failing outright once upstream added the provider-selection check
+    # above, with no non-interactive default. Pinning stops the next
+    # unannounced CLI change from breaking sandbox creation the same way.
+    - command: npm_config_legacy_peer_deps=true npx -y claude-mem@13.24.1 install --provider claude
       user: "1000"
-      description: Install latest claude-mem (plugin, marketplace registration, bun deps)
+      description: Install claude-mem 13.24.1 (plugin, marketplace registration, bun deps)
```
dvdksn: The version should be a kit arg I think

mdelapenya: You're right, and the blocker is on our side.

`SPEC-v2.md` §2.1 allows an arg anywhere in `spec.yaml`, substituted before the spec is decoded, so this is a legitimate position for one. Nothing in this repo implements that substitution though, so the literal text reaches the shell and the TCK job fails. Filed as docker/sbx-kits-contrib#278.

Keeping the version pinned here in the meantime — happy to revisit once args actually work in CI.

dvdksn (dismissed):

kiview (approved):

PR #282, opened 2026-09-10
fix(spec): gate LoadFromBytes on schemaVersion for a deterministic v1 contract

mdelapenya:
## Summary

Follow-up from the #178 review ([@ilopezluna's proposal](https://github.com/docker/sbx-kits-contrib/pull/178#pullrequestreview-4895200359)): `LoadFromBytes` accepted or rejected a v2 document depending on which keys it used. It now gates on `schemaVersion` before the v1 decode — `"1"` and omitted accepted, anything else rejected with a pointer to `LoadArtifactFromBytes`. A malformed document still falls through to the v1 decoder so its real syntax error surfaces.

Tests cover the full acceptance table and are mutation-checked; no callers outside the `spec` package exist.

@ilopezluna: keep or drop the `Deprecated:` marker? Kept as-is here — your call per #178.

Authored by Claude (Sonnet 5) on behalf of @mdelapenya.

### DIFF
--- a/spec/artifact.go
+++ b/spec/artifact.go
@@ -226,16 +226,10 @@ func LoadArtifactFromBytes(yamlBytes []byte) (*Artifact, error) {
 }
 
 // LoadFromBytes parses spec.yaml bytes into a normalized SpecFile using the v1
-// grammar only. Normalization (v1 sugar folding, legacy field removal) happens
-// inside this call.
-//
-// Unlike LoadArtifactFromBytes it does not fork on schemaVersion, so what happens
-// to a schemaVersion: "2" document depends on which keys it uses. Names the v1
-// grammar does not have (permissions, setup, agentInstructions) fail the strict
-// decode and return an error naming this limitation. Names v1 also has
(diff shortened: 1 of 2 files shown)


mdelapenya:
The e2e failures here are pre-existing on `main`, not introduced by this PR — the Kit TCK run for the #269 merge commit failed the same way on the same four kits ([run on main](https://github.com/docker/sbx-kits-contrib/actions/runs/34451367447)). The e2e layer #269 exported passes every kit via `--kit`, and current sbx builds reject that for `kind: sandbox` kits (`--kit ... must be kind "mixin"`); passing the kit's name positionally no longer helps either, because sbx now alias-resolves extracted agent names to the published kit instead of the local checkout. This PR touches `spec/`, which widens the e2e matrix to the agent kits and exposed it.

A kind-aware fix to the e2e invocation is coming as a separate PR; once it merges, this branch just needs a rerun.

Authored and posted by Claude.

kiview (approved):

PR #286, opened 2026-09-10
Add gitea kit

steilerDev:
## Summary

Adds `gitea/`, a `kind: mixin` kit that wires a Gitea access token into the sandbox proxy so an agent can use **git over HTTPS**, the **Gitea REST API**, and the **`tea` CLI** against a Gitea instance — self-hosted or `gitea.com` — with the token never entering the container.

The instance hostname is a kit argument (`--kit-arg gitea.host=git.example.com`, defaulting to `gitea.com`), so one kit serves any deployment.

**1. One inject rule covers git-over-HTTPS *and* the API — unlike the sibling `gitlab` kit.**

`gitlab/README.md` documents that it can only do API auth and pushes users to SSH for git. That limitation does not carry over, and the reason is on Gitea's side rather than the proxy's: `Authorization: token <t>` is handled by Gitea's `OAuth2` auth method, and Gitea registers its git-over-HTTP routes with that method enabled — [`routers/web/web.go`](https://github.com/go-gitea/gitea/blob/main/routers/web/web.go) passes `webAuth.AllowOAuth2` into `addOwnerRepoGitHTTPRouters`, under a comment noting sessionless auth exists precisely for "accessing git via http". So a single `header: Authorization` / `format: "token %s"` rule authenticates both surfaces, and no `gitea-ssh` companion kit is needed.

This has now been confirmed against a running self-hosted instance (see test plan), not just read out of the router source.

**2. Deliberately *not* using `scheme: basic`.**

`scheme: basic` plus `username` looks like the natural fit for git-over-HTTPS, but it is a decode-time sugar with nothing behind it. Normalization expands it to `Format: "%s"` and leaves `Header` empty; that empty header name is carried through into the proxy's service detector, which discards any auth config without one — so the request goes out **unauthenticated, silently**. `inject[].username` has no consumer. GitHub's Basic git auth is a hardcoded proxy special case for `github.com`, not something a kit can request.

Both the spec comments and the README say so, so the next author doesn't rediscover it the hard way. Arguably the spec should reject `scheme: basic` at validation time rather than accept it into a no-op — happy to file that separately if maintainers agree.

**3. `tea`'s login is seeded by writing its config file, not by `tea login add`.**

`tea login add` contacts the instance to validate the token, so it would 401 and fail `sbx create` whenever no credential is bound. The e2e run under `deny-all` exercises exactly that path and confirms the guard: the step exits 0 in 11ms with `mode=none` and sandbox creation proceeds.

The config is written with `printf` rather than a heredoc because the command is a YAML block scalar and a heredoc body would have to stay indented.

**4. Single inject entry is load-bearing, not incidental.** The engine reads `ApiKey.Inject[0]` and nothing else, so a second rule on the same domain would be silently ignored. Called out in a comment so nobody "helpfully" adds one.

## Test plan

`scripts/test-kit.sh` and `scripts/test-kit-e2e.sh` both pass locally on macOS/arm64. Since CI skips the e2e legs on fork PRs, the local e2e run is the only place those assertions have executed — it is green.

What the e2e run establishes beyond the standard assertions:

- **Network contract under `deny-all`**: the pinned `tea` download completes, so `dl.gitea.com` is the complete install-time egress set — nothing silently blocked.
- **Graceful no-credential path**: with no binding, the login-seeding step skips cleanly and `sbx create` still succeeds.
- `agentContext` lands at `kits-agent-context/gitea.md`.

Verified against a **live self-hosted instance** with a real token bound:

- `GITEA_TOKEN` inside the container holds the sentinel, not the token.
- `curl https://<host>/api/v1/user` authenticates with **no** auth header set by the client.
- `tea repos ls` authenticates — the proxy replaces tea's own `Authorization: token <sentinel>`.
- `git clone` of a **private** repo over HTTPS, and `git push`, both succeed with no credential helper, no SSH key, and no token in the remote URL. (A public repo was deliberately not used as the test: it would clone unauthenticated and prove nothing.)

Also checked while authoring:

- Arg substitution against a non-default host, including the `host:port` form (`git.example.com:3000`).
- `tea` 0.15.1 checksums verified out of band; the generated config parses (`tea login ls`, `tea login default` resolve).
- Captured the raw request `tea` emits against a local listener to confirm it sends `Authorization: token <token>` — i.e. exactly the header the inject rule targets, so no sentinel can leak upstream.

## Origin

Written against the repo's `spec/SPEC-v2.md` and the `kit-author` skill, using `gitlab/` as the structural template. The auth design came from reading Gitea's auth chain and the sbx proxy's credential-injection path rather than from trial and error.

---

### DIFF
--- a/gitea/README.md
+++ b/gitea/README.md
@@ -1,0 +1,86 @@
+# gitea — Gitea instance auth
+
+A mixin kit that wires a [Gitea](https://about.gitea.com/) access token into the sandbox proxy, so an agent can use **git over HTTPS**, the **Gitea REST API**, and the **`tea` CLI** against your instance — self-hosted or `gitea.com` — without the token ever entering the container. Pairs with any base agent (claude, codex, gemini, …).
+
+Gitea is not one of sbx's built-in auto-sign-on services (GitHub is), so out of the box a sandbox has no Gitea credentials and `sbx secret set gitea` alone has nothing to bind to. This kit closes that gap: it declares a `gitea` credential and injects it into outbound requests to your instance at the proxy. The container only ever sees a proxy-managed placeholder in `GITEA_TOKEN`.
+
+## Usage
+
+Store a Gitea access token once on the host. It needs the `write:repository` scope, plus `write:issue` if the agent should work with issues and pull requests:
+
+```console
+sbx secret set gitea
+```
+
+Then create a sandbox with the kit, naming your instance:
+
+```console
+sbx run --kit "docker.io/sbx/gitea-kit:latest" --kit-arg gitea.host=git.example.com claude
+```
+
+Or target this repo directly over git, or a local clone:
+
+```console
+sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=gitea" --kit-arg gitea.host=git.example.com claude
+sbx run --kit ./gitea/ --kit-arg gitea.host=git.example.com claude
+```
+
+`gitea.host` defaults to `gitea.com`, so you can drop `--kit-arg` entirely if that is the instance you use.
+
+Verify inside the sandbox:
+
+```console
+tea login ls                                              # already logged in
+curl -s https://git.example.com/api/v1/user               # no auth header needed
+git clone https://git.example.com/<owner>/<private-repo>.git
+```
+
+## How auth works
+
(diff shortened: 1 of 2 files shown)


kiview (approved):
Reviewed 8c391c942e390a64d13d3d5da90ab87b93b8cadd. No material correctness or security blockers found in the kit setup, host argument handling, or proxy-managed credential wiring. Checked the pinned tea checksums and login schema, the proxy injection path, and the passing container TCK logs. Full sandbox E2E jobs were skipped for this fork; the reported authenticated self-hosted tests were not independently rerun.

Authored and posted by Codex.

PR #287, opened 2026-09-10
Add Cloudsmith repository, CLI, and dependency firewall kits

BartoszBlizniak:
## Summary

Adds three composable mixins for using Cloudsmith as a sandbox's package source, with API access and public-registry blocking available separately:

- `cloudsmith-repo` configures pip/uv, npm/pnpm/Yarn, Go, Cargo, Maven, and NuGet for a repository selected through kit arguments. It also documents Docker and raw-download URLs and uses a proxy-injected entitlement token for private repositories.
- `cloudsmith-cli` installs the Cloudsmith CLI through a pinned, checksum-verified installer and adds a separate proxy-injected API key.
- `cloudsmith-dependency-firewall` denies the named public registries so clients cannot fall back to those endpoints.

Example from a local checkout:

```console
sbx secret set cloudsmith-entitlement-token
sbx run claude \
  --kit ./cloudsmith-repo/ \
  --kit ./cloudsmith-dependency-firewall/ \
  --kit-arg cloudsmith-repo.path=/your-org/your-repo .
```

Add `--kit ./cloudsmith-cli/` and store `cloudsmith-api-key` when API access is needed. Approve the relevant credential bindings on first use.

- First kits to use `args` (relies on the TCK substitution from #279) and `permissions.network.deny`.
- Repository credentials and API credentials are separate services. Package downloads need only a read-only, repository-scoped entitlement token; API access carries the permissions of the bound key.
- The repository kit defaults to Cloudsmith's public CLI repository so installation can be tested without credentials. Users select their own repository through `path` and can override the format hosts for custom domains.
- An install-time probe fails sandbox creation when the repository index does not answer 200. `sbx create` does not show the step's output, so the README points at the daemon log.
- The entitlement token is injected as `Authorization: Bearer`. `scheme: basic` is not usable with sbx 0.42.1 (it sends a bare token).
- The firewall enforces hostname rules. It does not restrict shared Cloudsmith hosts to one repository or block every alternative package source. The README describes those boundaries.
- The documented raw `versions/latest` redirect can expose an entitlement token in its response URL. The README retains this limitation, recommends version-pinned raw downloads, and explains rotation. Proxy injection must not be treated as a guarantee that hostile sandbox code cannot recover the token.
- Cargo, Maven, and NuGet home configuration files are replaced by the kit; this is documented as a prerequisite.

## Test plan

- [x] `sbx kit validate` and `sbx kit pack` for all three kits.
- [x] `go test ./spec ./tck` and `./scripts/test-kit.sh <kit>` for each kit, including container installation checks.
- [x] `./scripts/test-kit-e2e.sh <kit>` for each kit under `deny-all`, using the isolated `sbx-kits-contrib-tck` daemon. Every entry in `network.allow` came from `sbx policy log`.
- [x] Manual smoke with real credentials against a private repository on custom domains: pip, npm, Go, raw and Docker pulls succeed with no credential in the sandbox; quarantined versions return 403; with the firewall kit, the public registries return 403 and appear in `sbx policy log` as denied by the kit's rule; `cloudsmith whoami` authenticates through the proxy.

### DIFF
--- a/cloudsmith-cli/README.md
+++ b/cloudsmith-cli/README.md
@@ -7,7 +7,7 @@ Installs the [`cloudsmith` CLI](https://github.com/cloudsmith-io/cloudsmith-cli)
 Store the API key once (optional; without it the CLI installs and `cloudsmith whoami` reports no user):
 
 ```console
-sbx secret set cloudsmith-api-key -t <api-key>
+sbx secret set cloudsmith-api-key
 ```
 
 Create a sandbox, usually with the pull kit:
@@ -15,6 +15,8 @@ Create a sandbox, usually with the pull kit:
 ```console
 sbx run claude --kit "docker.io/sbx/cloudsmith-repo-kit:latest" --kit "docker.io/sbx/cloudsmith-cli-kit:latest" --kit-arg cloudsmith-repo.path=/acme/prod .
 ```
+
+The Git examples require `github.com/docker/` in Docker's `kit.allowedSources` setting. Preserve any existing allowed sources when adding it; see [Restrict kit sources](https://docs.docker.com/ai/sandboxes/customize/kits/#restrict-kit-sources).
 
 Or from git or a local clone:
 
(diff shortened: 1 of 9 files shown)


mdelapenya:
Hi @BartoszBlizniak, we merged https://github.com/docker/sbx-kits-contrib/pull/319 today. If you mind, I'll close this PR for that reason, but feel free to contribute to the merged one if you see any discrepancy. 

Cheers!

PR #291, opened 2026-09-11
Harden tck.yml's kit-detection step

jedp-docker:
## Summary
While porting this repo's validator into `sbx-kits-internal` (importing `spec`/`tck` and copying `tck.yml`'s structure), a code review pass on that copy caught three real bugs — all present here too, since they were copied verbatim:

- **`grep -vE` exits 1 on an empty `$changes` list** (e.g. a PR that only touches `README.md`). GitHub Actions runs `run:` blocks under `-e`/`pipefail`, so the failing pipeline aborts the "Collect kits to test" step before the `kits=${kits:-[]}` fallback ever runs — turning a harmless doc-only PR into a red CI run. Fixed with `|| true`.
- **`matrix.kit` interpolated directly into a `run:` script** — a kit directory name containing shell metacharacters could inject commands. `test-kit` has no fork-PR restriction (unlike the e2e jobs), so this is reachable from any PR. Fixed by passing it via `env:` instead.
- **Both kit-discovery call sites in `tck.yml` re-glob `spec.yaml`/`spec.yml`** instead of calling `scripts/discover-kits.sh`, despite that script's own comment claiming to be the single source of truth. It does have real callers elsewhere (`build-and-publish-kits.yml`, `hub-overview.yml`) — just not here. Both call sites now delegate to it.

While touching `discover-kits.sh`, two adjacent fixes:
- Anchored it to the repo root like its sibling scripts (`test-kit.sh`, `test-kit-e2e.sh`) already do — running it from any other cwd previously returned an empty list silently instead of erroring.
- Fixed the same `>> "$GITHUB_OUTPUT"` individual-redirects shellcheck nit (SC2129) in the neighboring "Generate filters" step.

Not in scope here: `actionlint` also flagged pre-existing, unrelated shellcheck issues in `build-and-publish-kits.yml` and `publish-one-kit.yml` (SC2129, SC2016) — left untouched since they're a different file and different concern from this PR.

## Test plan
- [x] `actionlint` clean on `tck.yml` (was previously flagging the SC2129 nit)
- [x] `discover-kits.sh` tested from the repo root and from a kit subdirectory; its two existing callers (`build-and-publish-kits.yml`, `hub-overview.yml`) unmodified
- [x] Reproduced the empty-`$changes` pipefail failure directly, confirmed the fix
- [x] `go build ./...` / `go vet ./...` pass (untouched by this change, sanity-checked anyway)
- [x] Live TCK run (`./scripts/test-kit.sh aider`) passes end to end
- [ ] CI: confirm `detect-changes` still selects the right kits on a real PR

### DIFF
--- a/.github/workflows/tck.yml
+++ b/.github/workflows/tck.yml
@@ -28,20 +28,22 @@ jobs:
         id: filter-setup
         run: |
           filters=""
-          # Add a filter for each kit directory (spec.yaml or spec.yml)
-          for f in */spec.yaml */spec.yml; do
-            [ -f "$f" ] || continue
-            kit=$(dirname "$f")
+          # Add a filter for each kit directory. discover-kits.sh is the
+          # single source of truth for "what is a kit" — delegate to it here
+          # too instead of re-globbing spec.yaml/spec.yml.
+          while IFS= read -r kit; do
             filters=$(printf '%s\n%s: "%s/**"' "$filters" "$kit" "$kit")
-          done
+          done < <(./scripts/discover-kits.sh)
           # Add filters for shared packages that should trigger all kits
           filters=$(printf '%s\ntck: "tck/**"\nspec: "spec/**"\nscripts: "scripts/**"' "$filters")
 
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):
LGTM, thanks!

PR #297, opened 2026-09-15
feat(gitlab,gitlab-ssh): support self-managed GitLab instances

johnfeeney-docker:
## Summary

The `gitlab` and `gitlab-ssh` mixins targeted gitlab.com only. This parameterizes both so one kit serves gitlab.com and any self-managed instance.

- `gitlab` gains `host` (default `gitlab.com`) and `service` (default `gitlab`). The host drives the network allow rule, the credential's inject domain, and `GITLAB_HOST`.
- `gitlab-ssh` gains `host` and `hostKey`, so a self-managed instance's SSH host key can be pinned the same way gitlab.com's published keys are.
- A second install hook seeds glab's host config, without which `glab auth status` reports a self-managed instance as unauthenticated even while every API call succeeds.
- Both defaults reproduce today's behaviour exactly, so an existing gitlab.com setup is unaffected and passes no arguments.

Before this change, a sandbox whose `gitlab` secret held a self-managed PAT could not be created **at all**: the kit's inject rule covered gitlab.com domain-wide, so the proxy attached that PAT to the install-time glab download. The download redirects to a gitlab.com `/api/v4/projects/…/packages/generic/…` path, and GitLab rejects a foreign token outright rather than falling back to anonymous access: `curl` exited 22 and the kit failed to apply.

- **gitlab.com stays allow-listed but is not an injection destination** unless it is the target host. That separation is what makes a self-managed target installable; a naive "add both domains to `inject`" fix reintroduces the 401 above and leaks an instance PAT to gitlab.com. Measured directly: that package path returns 200 anonymous and 401 with a bogus Bearer.
- **`service` is an argument** so a gitlab.com sandbox and a self-managed sandbox can hold different PATs under different service names. The alternative, per-sandbox secret scoping, collides at create time with `required: true` and is poor UX.
- **`hostKey` is explicit rather than scanned.** An install-time `ssh-keyscan` would be friendlier but pins whatever answers first, which is trust-on-first-use presented as pinning. A non-gitlab.com host with no `hostKey` fails the install loudly with instructions rather than producing a sandbox that cannot use SSH.
- **`GITLAB_HOST` is now always set, including for gitlab.com.** This pins the target rather than leaving glab to infer it from git remotes. Deliberate: a sandbox whose proxy injects a token for exactly one host should target that host deterministically.
- **The config-seeding hook writes no token** and is guarded by `[ ! -f ]`, since `setup.install` re-runs on recreate and glab writes to that same file. glab refuses to read a config that is not `0600`.
- **The self-managed path depends on `--kit-arg`, an experimental sbx flag.** gitlab.com users pass nothing and are unaffected.
- With `host` defaulting to gitlab.com the allow list contains `gitlab.com:443` twice. This validates clean; flagging it in case dedup is preferred.

## Test plan

CI covers spec validation and the TCK for both kits.

Run locally, all passing:

- `scripts/verify-kit-spec` on both kits: valid, no warnings.
- `scripts/test-kit.sh` on both kits: pass, including all three install hooks executing in a container. This is also what confirms the bumped glab digests resolve.
- `scripts/test-kit-e2e.sh` on both kits, under the scoped `deny-all` daemon: pass. Since the install hooks run under deny-all, this is what confirms the allow list is complete: the pinned glab tarball downloads with only the target host and `gitlab.com:443` listed.
- **gitlab.com with no arguments** (the defaults-unchanged check): sandbox creates, `GITLAB_TOKEN` stays the sentinel, `GITLAB_HOST=gitlab.com`, a bare `curl` to `/api/v4/user` returns 200 so the proxy injects, `glab auth status` reports gitlab.com authenticated, gitlab.com is in `known_hosts`, and SSH auth returns `Welcome to GitLab, @<user>!`.
- **Negative test, run for real**: `--kit-arg gitlab-ssh.host=<non-gitlab.com>` with no `hostKey` fails the install hook with exit 1, prints the guidance above, and creates no sandbox.
- Manual, against a real self-managed instance: sandbox creates; proxy injects (bare `curl` to `/api/v4/user` returns 200); `glab api user`, `glab repo view`, `issue list`, `mr list` all resolve; SSH auth and `git clone` over SSH succeed; `glab auth status` reports the instance. Clone over HTTPS fails as documented, since git sends Basic and the inject rule is Bearer.

## Origin

Raised while trying to use the existing `gitlab` kit against a self-managed instance, which surfaced the install-time failure above. Root-caused and developed against a live self-managed GitLab.


mdelapenya (approved):
LGTM, thanks!

PR #309, opened 2026-09-17
fix(kernel): refresh kit

rgarcia:
## Summary

- mark the Kernel API key as required and add source and license metadata
- add agent instructions that point to the bundled quick reference and require browser-session cleanup
- update credential setup and usage docs to lead with the published Docker Hub artifact
- fix the replay stop parameter and distinguish browser profiles from managed auth in the quick reference

- Credential injection remains restricted to `api.onkernel.com`; the broader `*.onkernel.com` network allowance is still needed for regional CDP WebSocket endpoints and does not receive auth injection.
- `sourceURL` points to the upstream `kernel/` directory because this repository is the source of the published OCI artifact.
- The existing package registry and GitHub release hosts remain in the allow-list because the install hook and optional SDK installs need them under `deny-all`.

## Test plan

- [x] `sbx kit validate ./kernel/`
- [x] `./scripts/test-kit.sh kernel`
- [x] Type-checked the TypeScript quick-reference examples against `@onkernel/sdk@0.107.0`
- [x] Parsed the Python quick-reference example
- [ ] `./scripts/test-kit-e2e.sh kernel` — the fork's scoped `sbx` daemon is not authenticated to Docker Hub; per the repository docs, fork PR CI also skips e2e because it cannot access the required secrets
- [ ] `sbx run --kit ./kernel/ claude` — blocked by the same Docker Hub login prerequisite

## Origin

Ported from the maintained [`kernel/docker-sbx-kit`](https://github.com/kernel/docker-sbx-kit) source after merging [`kernel/docker-sbx-kit#1`](https://github.com/kernel/docker-sbx-kit/pull/1).


mdelapenya (approved):
LGTM, thanks!

PR #310, opened 2026-09-18
Add tessl kit

sjmaple:
## Summary

A mixin that installs the Tessl CLI, authenticates it via proxy-injected API-key auth, and wires the Tessl MCP server into the sandboxed agent so it can search and install skills and tiles from the Tessl registry. Pairs with any base agent — `tessl init` writes MCP config for eight agents and each reads only its own file.

Ported from an existing kit authored against `schemaVersion: "1"`. Beyond the mechanical v1 → v2 migration (`commands` → `setup`, `network.allowedDomains` → `permissions.network.allow`, the four credential blocks → a unified `credentials[]`, and the `CLAUDE.md` append hook → `agentInstructions`), testing against a live sandbox turned up three bugs that were silently shipping in the original:

- **Auth never worked.** `apiKey.proxyManaged` sets `TESSL_TOKEN` to the literal `proxy-managed`, but the CLI shape-checks the variable client-side and rejects anything not starting with `sk_`, so it reported "not logged in" and never issued a request for the proxy to inject into. The placeholder is now set via `environment.variables` as `sk_proxy_managed`. Injection is keyed on the destination domain, not the variable's value, so the security properties are unchanged.
- **The MCP wiring never happened.** `--agent agents` is not a valid value in tessl 0.109.0, and one bad value aborts the whole `tessl init`, writing no config at all — so the kit's headline feature was dead.
- **Config was written to the wrong directory.** A startup hook's `PWD` is `/home/agent/workspace`, which is the real working directory only when the workspace was cloned in; under a direct mount the workspace is bind-mounted at its host path. `tessl init` reported success while writing where the agent would never look. The hook now `cd`s to `$WORKSPACE_DIR`.

- **`TESSL_TOKEN` is set in `environment.variables`, not with `apiKey.proxyManaged: true`.** This looks like a mistake and isn't — see the first bug above. The sentinel value isn't configurable per-credential (`proxyManaged` hardcodes `proxy-managed`, unlike `oauth.sentinels`), so a placeholder shaped to pass the CLI's check has to come from `environment.variables`. An `apiKey.placeholder` field upstream would let this kit use `proxyManaged` as intended.
- **The install is `curl | sh`, not a version+digest-pinned tarball.** Deliberate: the installer fetches a signed `SHA256SUMS` manifest, verifies its ECDSA P-256 signature against an embedded public key, and checks the tarball hash before extracting, on by default and failing closed. Hand-copying a digest into `spec.yaml` would replace a signature check with a weaker one. The version is pinned via a `version` kit arg exported as `TESSL_VERSION`.
- **`permissions.network.allow` is a wildcard (`*.tessl.io:443`) plus the apex.** Not laziness — the registry serves some dynamically-named subdomains (a `7530.tessl.io` showed up in `policy log` during e2e), so an enumerated host list breaks `tessl search`. The apex needs its own rule since `*.` matches exactly one label, never zero.
- **`ports.ubuntu.com` and `download.docker.com` are deliberately *not* allowlisted**, though they appear as blocked in `policy log`. They come from the base agent kit's own best-effort `apt-get update`, which is guarded with `|| true` and fails soft. This mixin is base-agnostic and its own install touches no apt source, so those domains belong to whichever sandbox kit it's composed with — compare `ecc`, which does list them because it pins itself to `claude` via `requires.agent`.
- **`TESSL_TELEMETRY=off`** so the egress contract stays `*.tessl.io` rather than needing an analytics host allowlisted.

## Origin

Ported from [shelajev/tessl-sbx-kit](https://github.com/shelajev/tessl-sbx-kit) (Apache-2.0), which was authored against `schemaVersion: "1"`; credited in `sourceURL` and the kit README.

## Test plan

CI runs `kit validate` and the TCK on every PR. CI does **not** run e2e on

- [x] `sbx kit validate ./tessl/` passes — directory and packed zip, no warnings from `sbx kit inspect --json`
- [x] `./scripts/test-kit.sh tessl` passes (the TCK) — including the `container` leg, with both install hooks executing
- [x] `./scripts/test-kit-e2e.sh tessl` passes. The script applies the same
      `deny-all` baseline CI uses and scopes everything to its own daemon
      (`--app-name sbx-kits-contrib-tck`), so my main sbx state is untouched.
      Every entry I added to `permissions.network.allow` came from
      `sbx --app-name sbx-kits-contrib-tck policy log <tck-e2e-…>`, not a guess.
- [x] Manual smoke: `sbx run --kit ./tessl/ claude` and verified the kit's
      binary / files / env are inside the running container.

The manual smoke went further than the checklist, since the three bugs above all pass `validate`, the TCK and e2e while leaving the kit non-functional. In a live sandbox: `tessl whoami` authenticating through proxy injection, `tessl search` returning results, `tessl install` completing under `deny-all` with no new blocked hosts, an MCP `initialize` handshake answered by the Tessl MCP server (28 tools), `.mcp.json` and `tessl.json` landing in the mounted workspace, `CLAUDE.md`/`AGENTS.md` appended rather than overwritten, and restart idempotency across a stop/start.

**One gap worth knowing:** all of the above ran on **arm64** (Apple Silicon); CI runs amd64. The install hook is the cross-arch risk — the Tessl installer does its own musl and architecture detection, which nothing here has exercised on amd64. Happy to rerun if a maintainer can point me at an amd64 path, or to treat CI's TCK leg as the check.


tessl/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Mint an API key on the host and bind it once, then run with the kit attached:
+
+```console
+# On the host, in a shell logged in to Tessl. --workspace, --name and --role
+# are all required; `tessl workspace list` shows the workspaces you belong to.
+tessl api-key create --workspace <workspace> --name sbx --role member
+
+sbx secret set tessl      # paste the key when prompted
+sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=tessl" claude
```
mdelapenya, line 17: suggestion: recommend the OCI format before the git one.
```suggestion
sbx run --kit "docker.io/sbx/tessl-kit:latest" claude
```

tessl/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+Keys expire 30 days from creation unless you pass `--expiry-date`.
+
+Or from the published OCI artifact:
+
+```console
+sbx run --kit "docker.io/sbx/tessl-kit:latest" claude
```
mdelapenya, line 25:
```suggestion
Or from the git repository:

```console
sbx run --kit "git+https://github.com/docker/sbx-kits-contrib.git#dir=tessl" claude
```

PR #314, opened 2026-09-18
Add mem0 kit

ajeetraina:
## Add `mem0` kit 

A `kind: mixin` (schemaVersion 2) kit that adds the [Mem0](https://github.com/mem0ai/mem0) memory layer to any agent sandbox, pre-wired to a local [Docker Model Runner](https://docs.docker.com/ai/model-runner/) for both the LLM and the embedder. No cloud credentials, no external vector database.

### What it does
- Installs `mem0ai[nlp]==2.0.5` + `click` (pinned) and the spaCy `en_core_web_sm` model.
- Writes `~/.mem0/config.json` (only if missing) wiring Mem0 to `ai/gemma3` (LLM) and `ai/mxbai-embed-large` (embedder) on DMR, with an on-disk Qdrant vector store.
- Reaches DMR at `host.docker.internal:12434`, routed around the proxy via `NO_PROXY`.

### Design notes
- **Agent-agnostic**: Mem0 is a Python library the agent calls, not an agent-specific hook, so no `requires: agent:` affinity.
- **DMR-only, by design**: this variant needs no cloud key. OpenAI and Gemini backends differ in egress, env vars, pip deps, and the whole `config.json`, so they belong in sibling kits (`mem0-openai`, `mem0-gemini`) following the `claude-acp`/`codex-acp` per-backend precedent, rather than a single arg-switched spec.

### Validation
- `sbx kit validate ./mem0/` -> VALID
- `schemaVersion: "2"`, `kind: mixin`, no args, config.json materialized via `setup.files`.


mem0/README.md
```diff
@@ -0,0 +1,56 @@
+# mem0 - Mem0 memory layer (Docker Model Runner)
```
mdelapenya, line 1: question: should be rename this folder to `mem0-dockermodelrunner` so that we leave space for a future mem0 contribution?

PR #317, opened 2026-09-18
Add PANW kits

ajeetraina:
## Add PANW kits

Two `kind: mixin` (schemaVersion 2) kits from the Palo Alto Networks integration. The source published them as two tags in one Hub repo; in the one-kit-per-directory model they become two directories, added together here (following the `claude-acp`/`codex-acp` precedent from #151).

### `endpoint-enforcement`
Marks agent processes as sandbox-wrapped (`SANDBOX_ENFORCED=1` + a marker id env var + a read-only `~/.sandbox-enforced` marker file) so a host-side endpoint policy can permit sandbox-wrapped agents and deny agents spawned outside a sandbox. `markerId` is a kit arg. No network egress, no secret.

### `siem-telemetry`
Background Fluent Bit forwarder shipping sandbox logs (`/var/log/sandbox/`, `~/.sandbox/logs/`) to a SIEM HTTP event collector (targets Cortex XSIAM, works with any HTTP collector).
- **Install by page size**: prebuilt Fluent Bit on 4KB-page hosts; source build with `FLB_JEMALLOC=Off` on 16KB-page hosts (Apple-Silicon microVM), where the prebuilt jemalloc aborts.
- **Auth**: the collector token is a custom secret (`sbx secret set-custom`, host-keyed); the sandbox only sees a placeholder that the proxy overwrites on the wire. The non-secret XSIAM `x-xdr-auth-id` is a kit arg.
- `siemCollectorHost` / `siemCollectorPath` / `siemCollectorAuthId` are kit args, substituted into the fluent-bit config shipped under `files/home/.config/fluent-bit/`.

### Notes
- `siemCollectorHost` was `required: true` with no default in the source; changed to a placeholder default so `sbx kit validate` (run with no args by `publish-artifact.sh`) and the arg-default TCK pass.
- Excluded the source's manual `test/` harness (mock collector) and repo-level docs; kept only kit runtime files.

### Validation
- `sbx kit validate` -> VALID for both, with and without `--kit-arg`.
- Arg substitution confirmed (marker env/file for endpoint-enforcement; collector host/path/auth-id into the fluent-bit config for siem-telemetry).


endpoint-enforcement/README.md
```diff
@@ -0,0 +1,38 @@
+# endpoint-enforcement - sandbox enforcement marker
```
mdelapenya: question: should we rename this folder to `panw-endpoint-enforcement`, or `panw-ef`?

siem-telemetry/README.md
```diff
@@ -0,0 +1,61 @@
+# siem-telemetry - forward sandbox telemetry to a SIEM
```
mdelapenya: question: should we rename this folder to `panw-siem-telemetry`, or `panw-st`?

ajeetraina: I would suggest `panw-siem-telemetry` as it's clearer, more discoverable name and matches the "prefix by vendor" convention.

mdelapenya (approved):

rcjsuen (approved):

PR #320, opened 2026-09-18
Add surrealdb kit

ajeetraina:
## Add `surrealdb` kit 

A `kind: mixin` kit that adds an embedded, multi-model SurrealDB (documents, graph, native vector search) plus the `surrealdb` Python SDK to any agent, with vector search pre-wired to a local Docker Model Runner embedder. No cloud credentials, no separate database server. Migrated from schemaVersion 1 to 2.

### What it does
- Installs `surrealdb==2.0.0` (embedded storage engines bundled) + `openai>=1.0` (to call the DMR embedder).
- Writes `~/.surrealdb/config.json` (only if missing) wiring the surrealkv store and the `ai/mxbai-embed-large` embedder on DMR (1024 dims).
- Ships runnable demos under `files/home/runbooks/` (-> `~/runbooks/`): `travel.py` (vector memory), `graph.py` (multi-model graph).
- Agent-agnostic; DMR is reached at `host.docker.internal:12434`, routed around the proxy via `NO_PROXY`.

### Validation
- `sbx kit validate ./surrealdb/` -> VALID (schemaVersion 2).


surrealdb/files/home/runbooks/graph.py
```diff
@@ -0,0 +1,55 @@
...
+db = Surreal(cfg["db"]["url"])
+db.connect()
+try:
+    db.signin({"username": "root", "password": "root"})
+except Exception:
+    pass  # embedded engines don't authenticate
+db.use(cfg["db"]["namespace"], "graph_demo")
+
+# Re-runnable: start from a clean slate each time.
+for t in ("person", "city", "visited"):
```
mdelapenya, line 29: question: is this needed in the kit?

ajeetraina: It keeps the runbook re-runnable. The records use fixed IDs (`person:alice`, etc.), and SurrealDB's `CREATE` errors if the record already exists, so without this loop the demo works once and a second run fails with `person:alice already exists`. The `REMOVE TABLE IF EXISTS` wipes the demo tables so each run starts clean. I'll add a comment making that explicit.

